PR #6, opened 2018-10-29
Moddataimp 33

OleksiiKuzminov:
* Design service methods
* Add all methods to interface and javadocs
* Use methods calls at API endpoints

### DIFF
--- a/src/main/java/org/folio/dao/UploadDefinitionDao.java
+++ b/src/main/java/org/folio/dao/UploadDefinitionDao.java
@@ -1,0 +1,56 @@
+package org.folio.dao;
+
+import io.vertx.core.Future;
+import org.folio.rest.jaxrs.model.UploadDefinition;
+import org.folio.rest.jaxrs.model.UploadDefinition;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Data access object for {@link UploadDefinition}
+ */
+public interface UploadDefinitionDao {
+
+  /**
+   * Searches for {@link UploadDefinition} in database
+   *
+   * @param query  CQL query
+   * @param offset offset
+   * @param limit  limit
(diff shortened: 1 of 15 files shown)


KaterynaSenchenko:
Shouldn't it be v1.0 [https://github.com/folio-org/mod-data-import/blob/60dd8fb07e0e568dcabf39bd87605ef8454fb12a/ramls/dataImport.raml#L4](https://github.com/folio-org/mod-data-import/blob/60dd8fb07e0e568dcabf39bd87605ef8454fb12a/ramls/dataImport.raml#L4)? Otherwise, looks good to me.

src/main/java/org/folio/dao/UploadDefinitionDao.java
```diff
@@ -0,0 +1,56 @@
...
+   */
+  Future<Optional<UploadDefinition>> getUploadDefinitionById(String id);
+
+  /**
+   * Saves {@link UploadDefinition} to database
+   *
+   * @param UploadDefinition {@link UploadDefinition} to save
+   * @return future with id of saved {@link UploadDefinition}
+   */
+  Future<String> addUploadDefinition(UploadDefinition UploadDefinition);
```
KaterynaSenchenko: Please start parameter name with lower-case letter

src/main/java/org/folio/dao/UploadDefinitionDao.java
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Updates {@link UploadDefinition} in database
+   *
+   * @param UploadDefinition {@link UploadDefinition} to update
+   * @return future with true is succeeded
+   */
+  Future<Boolean> updateUploadDefinition(UploadDefinition UploadDefinition);
```
KaterynaSenchenko: same as above

PR #7, opened 2018-11-01
Upload Definition Service Implimentation

OleksiiKuzminov:
add Upload definition services impl
change file API


src/main/java/org/folio/service/LocalFileStorageService.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class LocalFileStorageService implements FileStorageService{
+  @Override
+  public String getServiceName() {
+    return "LOCAL_STORAGE";
+  }
+
+  @Override
+  public Future<File> getFile(String path) {
+    return null;
```
Igor-Gorchakov: Please left TODO regard methods with stub impls

src/main/java/org/folio/service/FileServiceImpl.java
```diff
@@ -16,45 +16,56 @@
...
+  @Override
+  public Future<String> addFile(File file) {
     file.setId(UUID.randomUUID().toString());
-    return FileDao.addFile(file);
+    return fileDao.addFile(file);
+  }
+
+  @Override
+  public Future<String> uploadFile(String fileId, DataImportUploadFileFileIdPostMultipartFormData data) {
+//    fileDao.getFileByUploadDefinitionId(file.getUploadDefinitionId()).map(files -> files.);
```
Igor-Gorchakov: TODO

src/main/java/org/folio/service/FileServiceImpl.java
```diff
@@ -16,45 +16,56 @@
...
   }
 
   @Override
-  public Future<String> addFile(File file, DataImportUploadFilePostMultipartFormData data) {
+  public Future<List<File>> getFileByUploadDefinitionId(String id) {
+    return fileDao.getFileByUploadDefinitionId(id);
+  }
+
+  @Override
+  public Future<String> addFile(File file) {
```
Igor-Gorchakov: Will this service be able to add multiple list of files at one call ?
Or the user of this interface should call addFile method per each batch upload ?

src/main/java/org/folio/service/file/FileServiceImpl.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  public FileServiceImpl(Vertx vertx, String tenantId) {
+    this.vertx = vertx;
+    this.tenantId = tenantId;
+    this.uploadDefinitionService = new UploadDefinitionServiceImpl(vertx, tenantId);
+  }
+
+  @Override
+  public Future<UploadDefinition> uploadFile(String fileId, String uploadDefinitionId, InputStream data, Map<String, String> okapiHeaders) {
```
OleksiiKuzminov: and how do you like this, Elon Musk?

PR #9, opened 2018-11-20
Moddataimp 41

OleksiiKuzminov:
Changes:
- add ret call to mod-source-record-manager
- replace jobExecution ids from a response
- add mocks for API tests
-bugfixing


src/main/java/org/folio/dao/UploadDefinitionDaoImpl.java
```diff
@@ -25,13 +26,108 @@
...
   public UploadDefinitionDaoImpl(Vertx vertx, String tenantId) {
     pgClient = PostgresClient.getInstance(vertx, tenantId);
+    this.schema = PostgresClient.convertToPsqlStandard(tenantId);
+  }
+
+  public interface UploadDefinitionMutator {
+    Future<UploadDefinition> mutate(UploadDefinition definition);
+  }
+
+  public Future<UploadDefinition> updateBlocking(String uploadDefinitionId, UploadDefinitionMutator mutator) {
```
OleksiiKuzminov: plz take a look on this code

PR #10, opened 2018-11-21
Moddataimp 41

OleksiiKuzminov:
Changes:

add ret call to mod-source-record-manager
replace jobExecution ids from a response
add mocks for API tests
-bugfixing
add blocking update mechanism

### DIFF
--- a/src/main/java/org/folio/dao/UploadDefinitionDaoImpl.java
+++ b/src/main/java/org/folio/dao/UploadDefinitionDaoImpl.java
@@ -16,6 +16,7 @@ import org.folio.rest.persist.cql.CQLWrapper;
 import org.folio.rest.persist.interfaces.Results;
 import org.z3950.zing.cql.cql2pgjson.CQL2PgJSON;
 
+import javax.ws.rs.NotFoundException;
 import java.util.Optional;
 
 public class UploadDefinitionDaoImpl implements UploadDefinitionDao {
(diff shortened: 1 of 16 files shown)


src/main/java/org/folio/dao/UploadDefinitionDaoImpl.java
```diff
@@ -25,13 +26,116 @@
...
+  @FunctionalInterface
+  public interface UploadDefinitionMutator {
+    /**
+     * @param definition - Loaded from DB UploadDefinition
+     * @return - changed Upload Definition ready for save into database
+     */
+    Future<UploadDefinition> mutate(UploadDefinition definition);
+  }
+
+  public Future<UploadDefinition> updateBlocking(String uploadDefinitionId, UploadDefinitionMutator mutator) {
```
OleksiiKuzminov: plz, take a look on this update method

src/main/java/org/folio/service/file/FileServiceImpl.java
```diff
@@ -32,30 +33,32 @@ public FileServiceImpl(Vertx vertx, String tenantId, UploadDefinitionService upl
...
-                    .withFileDefinitions(replaceFile(def.getFileDefinitions(), savedFile))
-                    .withStatus(UploadDefinition.Status.IN_PROGRESS))
-                ).orElseThrow(NotFoundException::new))))
-        )
-        .orElseThrow(NotFoundException::new))
-      .compose(reply -> reply)
-      .compose(fileReply -> fileReply)
-      .compose(defReply -> defReply)
-      .compose(result -> result);
+  public Future<UploadDefinition> uploadFile(String fileId, String uploadDefinitionId, InputStream data, OkapiConnectionParams params) {
```
OleksiiKuzminov: and how it works now

src/main/java/org/folio/dao/UploadDefinitionDaoImpl.java
Igor-Gorchakov: Maybe updateExclusive ? , coz this is exclusive mode, which means the second select cannot proceed until the first one has completed or rolled back.

src/main/java/org/folio/service/upload/UploadDefinitionService.java
```diff
@@ -47,6 +49,15 @@
    */
   Future<UploadDefinition> updateUploadDefinition(UploadDefinition uploadDefinition);
 
+  /**
+   * Updates {@link UploadDefinition} in database with row blocking
```
Igor-Gorchakov: Please add java doc about exclusive mode, which means the second select cannot be executed until the first one has completed or rolled back.

PR #28, opened 2019-01-16
Moddataimp 45 reader impl

Igor-Gorchakov:
Implement reading and splitting source records. The process should run in the background.


src/main/java/org/folio/service/processing/reader/LocalStorageSourceReader.java
```diff
@@ -0,0 +1,117 @@
...
+    RecordsBuffer recordsBuffer = new RecordsBuffer();
+    /* Reading source data to the RecordsBuffer while it is not full and the file has not come to the end. */
+    while (!recordsBuffer.isFull() && this.hasNext) {
+      CharBuffer readBuffer = CharBuffer.allocate(READ_BUFFER_SIZE);
+      try {
+        while (this.inputStreamReader.read(readBuffer) > 0) {
+          readBuffer.flip();
+          this.hasNext = readBuffer.remaining() == readBuffer.capacity();
+          String source = this.sourceCache.append(readBuffer).toString();
+          readBuffer.clear();
```
OleksiiKuzminov: maybe sourceCache clean?

OleksiiKuzminov:
plz, make sonar happy
need tests for all functionality. Without test, we can't say that task is ready

src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
```diff
@@ -116,33 +112,41 @@ public void process(JsonObject jsonRequest, JsonObject jsonParams) {
...
+    int totalChunkCounter = 0;
+    try {
+      SourceReader reader = new LocalStorageSourceReader(fileStorageService.getFile(fileDefinition.getSourcePath()));
+
+      /*
+        If one of the dedicated handlers for sending chunks is failed, then all the other senders have to be aware of that
+        in a terms of the target file processing.
+      */
+      AtomicBoolean canSendNextChunk = new AtomicBoolean(true);
+      for (List<String> records = reader.readNext(); records.size() > 0; records = reader.readNext(), totalChunkCounter++) {
```
OleksiiKuzminov: change to total tecords

src/main/java/org/folio/service/processing/reader/RecordsBuffer.java
```diff
@@ -0,0 +1,32 @@
...
+  public List<String> getRecords() {
+    return this.records;
+  }
+
+  public void add(List<String> records) {
+    this.records.addAll(records);
+  }
+
+  public boolean isFull() {
+    return this.records.size() > CHUNK_SIZE;
```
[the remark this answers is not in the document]
OleksiiKuzminov: />=

src/main/java/org/folio/service/processing/reader/LocalStorageSourceReader.java
```diff
@@ -0,0 +1,117 @@
...
+      String errorMessage = String.format("Can not close stream reader for the file: %s. Cause: %s.", this.file.getPath(), e.getCause());
+      LOGGER.error(errorMessage);
+      throw new IllegalStateException(errorMessage);
+    }
+  }
+
+  private void initInputStreamReader() {
+    if (this.inputStreamReader == null) {
+      try {
+        InputStream inputStream = new FileInputStream(this.file);
```
OleksiiKuzminov: openInputStream from Apache IO

src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
KaterynaSenchenko: I think, 'totalRecords' would be a misleading name since only the last chunk with no records contains an actual total, for all the other chunks it's more like a counter, so maybe 'recordsCounter' is a better option? In this case, 'recordsCounter' value for the last chunk would be equal to total.

src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
```diff
@@ -116,33 +112,41 @@ public void process(JsonObject jsonRequest, JsonObject jsonParams) {
...
+        /* Sending the next read chunk */
+        if (canSendNextChunk.get()) {
+          RawRecordsDto chunk = new RawRecordsDto().withLast(false).withRecords(records).withTotal(totalChunkCounter);
```
KaterynaSenchenko: Probably we should update RawRecordsDto data model - total is not really a total number of records for all the chunks except the last one. Definitely, the description of 'total' field should be updated and maybe the field name itself, e.g. 'counter', what do you think?

KaterynaSenchenko:
Could you use validateAsyncResult method from RestUtils here https://github.com/folio-org/mod-data-import/pull/28/files#diff-051e96d497fc9fba8c8dd6cc6063166eL160? If it doesn't contain validation of 204 response code, maybe we should add it there.

src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
```diff
@@ -114,33 +110,46 @@ public void process(JsonObject jsonRequest, JsonObject jsonParams) {
...
+          if (ar.failed()) {
+            canSendNextChunk.set(false);
+            String errorMessage =
+              "File processing stopped. Can not send chunks of the file " + fileDefinition.getSourcePath() + " to the consumer.";
+            LOGGER.error(errorMessage);
+            future.fail(errorMessage);
+          }
+        });
+      }
+      if (!future.failed()) {
```
OleksiiKuzminov: remove this

PR #36, opened 2019-02-05
MODDATAIMP-82 Removed JobExecution status update during file processing

KaterynaSenchenko:


src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
```diff
@@ -33,11 +33,9 @@
...
 import java.util.concurrent.atomic.AtomicBoolean;
 
-import static org.folio.dataImport.util.RestUtil.validateAsyncResult;
+import static org.folio.dataimport.util.RestUtil.validateAsyncResult;
 import static org.folio.rest.RestVerticle.MODULE_SPECIFIC_ARGS;
 import static org.folio.rest.jaxrs.model.StatusDto.Status.ERROR;
-import static org.folio.rest.jaxrs.model.StatusDto.Status.IMPORT_FINISHED;
-import static org.folio.rest.jaxrs.model.StatusDto.Status.IMPORT_IN_PROGRESS;
 
 /**
```
Igor-Gorchakov: Please add java doc about case when job execution gets ERROR status

PR #37, opened 2019-02-08
upload changed to the stream mode

OleksiiKuzminov:


src/main/java/org/folio/service/storage/LocalFileStorageService.java
```diff
@@ -58,7 +54,7 @@ public String getServiceName() {
                 logger.error("Error during calculating path for file save. FileId: {}", fileId);
                 future.fail(r.cause());
               } else {
-                logger.info("File was saved to the storage. File size {} mb", (r.result() / 1024) / 1024);
+                logger.info("File part was saved to the storage.");
```
Igor-Gorchakov: Please log fileDefinitionId also

src/main/java/org/folio/rest/impl/DataImportImpl.java
```diff
@@ -187,15 +190,24 @@ public void deleteDataImportUploadDefinitionFileByFileId(String fileId, String u
...
     vertxContext.runOnContext(c -> {
       try {
+        byte[] data = IOUtils.toByteArray(entity);
         OkapiConnectionParams params = new OkapiConnectionParams(okapiHeaders, vertxContext.owner());
-        fileService.uploadFile(fileId, uploadDefinitionId, entity, params)
-          .map(PostDataImportUploadFileResponse::respond200WithApplicationJson)
+        if (futureforUplaod == null) {
+          futureforUplaod = fileService.beforeFileSave(fileId, uploadDefinitionId, params);
+        }
+        futureforUplaod = futureforUplaod.compose(def ->
```
Igor-Gorchakov: please choose better naming for "futureforUplaod"

PR #38, opened 2019-02-14
MODDATAIMP-81 Updated API

KaterynaSenchenko:
* renamed endpoints
* moved CRUD endpoints for FileExtensions from mod-source-record-storage
* refactored tests


OleksiiKuzminov:
Hange modulePermissions in moduledescriptor. 
Also need to add tenant api with setuping file extensions

PR #46, opened 2019-02-26
Fixed updating JobExecution status when UploadDefinition has been deleted

RuslanLavrov:


src/main/java/org/folio/service/upload/UploadDefinitionServiceImpl.java
```diff
@@ -108,17 +108,16 @@ public UploadDefinitionServiceImpl(Vertx vertx, String tenantId) {
...
-                    }
-                  }
-                );
+                .compose(result -> {
+                  updateJobExecutionStatuses(jobExecutionList, new StatusDto().withStatus(StatusDto.Status.DISCARDED), params)
+                    .otherwise(throwable -> {
+                      logger.error("Couldn't update JobExecution status to DISCARDED after UploadDefinition {} was deleted", id, throwable);
+                      return result;
+                    });
+                  Future<Boolean> future = Future.future();
```
Igor-Gorchakov: just return Future.succeededFuture(result)

PR #65, opened 2019-04-08
Added update uploadDefinition status after files processing

RuslanLavrov:


src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
```diff
@@ -75,20 +76,28 @@ public void process(JsonObject jsonRequest, JsonObject jsonParams) { //NOSONAR
...
                     } else {
                       LOGGER.info("File {} successfully processed.", fileDefinition.getSourcePath());
                       blockingFuture.complete();
                     }
                   }),
                 false,
-                null
+                blockingAr -> {
+                  processFilesFutures.add(blockingAr.succeeded() ? Future.succeededFuture() : Future.failedFuture(blockingAr.cause()));
+                  if (processFilesFutures.size() == fileDefinitions.size()) {
```
OleksiiKuzminov: else?

PR #68, opened 2019-04-12
Moddataimp 116 - Update fileDefinition status on upload fail

RuslanLavrov:


src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
```diff
@@ -90,19 +89,11 @@ public void process(JsonObject jsonRequest, JsonObject jsonParams) { //NOSONAR
...
-                    .map(e -> uploadDefinitionService.updateBlocking(uploadDefinition.getId(), definition -> Future.succeededFuture(uploadDefinition.withStatus(COMPLETED)), tenantId))
-                    .otherwise(uploadDefinitionService.updateBlocking(uploadDefinition.getId(), definition -> Future.succeededFuture(uploadDefinition.withStatus(UploadDefinition.Status.ERROR)), tenantId));
-                  }
-                }
+                null
               );
             }
             return Future.succeededFuture();
-          });
+          }).setHandler(event -> uploadDefinitionService.updateBlocking(uploadDefinition.getId(), definition -> Future.succeededFuture(definition.withStatus(COMPLETED)), tenantId));
```
OleksiiKuzminov: format code plz to more readable view

src/main/java/org/folio/service/upload/UploadDefinitionServiceImpl.java
```diff
@@ -378,6 +380,16 @@ private Errors validateFreeSpace(UploadDefinition definition, Errors errors) {
     }, tenantId);
   }
 
+  @Override
+  public Future<Boolean> updateUploadDefinitionStatusToError(UploadDefinition uploadDefinition, String tenantId) {
```
OleksiiKuzminov: bad practice. Use status as a method argument

Igor-Gorchakov:
Please update news.md

src/main/java/org/folio/service/upload/UploadDefinitionServiceImpl.java
```diff
@@ -378,6 +380,16 @@ private Errors validateFreeSpace(UploadDefinition definition, Errors errors) {
...
+  @Override
+  public Future<Boolean> updateUploadDefinitionStatusIfAllFilesUploadingFailed(UploadDefinition uploadDefinition, UploadDefinition.Status status, String tenantId) {
+    return uploadDefinition.getFileDefinitions()
+      .stream()
+      .allMatch(fileDefinition -> fileDefinition.getStatus().equals(ERROR))
+        ? uploadDefinitionDao.updateBlocking(uploadDefinition.getId(), definition ->
+          Future.succeededFuture(definition.withStatus(UploadDefinition.Status.ERROR)), tenantId)
+          .map(true)
+        : Future.succeededFuture(false);
+  }
```
Anatolii-Starkov: @RuslanLavrov Don't go crazy. I can't read it 😄 
I suggest the following
```suggestion
 public Future<Boolean> updateUploadDefinitionStatusIfAllFilesUploadingFailed(
          UploadDefinition uploadDefinition,
          Status status,
          String tenantId) {

    Future<Boolean> result = Future.succeededFuture(false);

    if (areAllDefinitionsFailed(uploadDefinition)) {
      result = uploadDefinitionDao
              .updateBlocking(uploadDefinition.getId(), it -> Future.succeededFuture(it.withStatus(Status.ERROR)), tenantId)
              .map(true);
    }

    return result;
  }


  private boolean areAllDefinitionsFailed(UploadDefinition uploadDefinition) {

    return uploadDefinition.getFileDefinitions().stream()
            .allMatch(it -> it.getStatus() == FileDefinition.Status.ERROR);
  }
```

src/main/java/org/folio/service/upload/UploadDefinitionServiceImpl.java
```diff
@@ -378,6 +380,22 @@ private Errors validateFreeSpace(UploadDefinition definition, Errors errors) {
     }, tenantId);
   }
 
+  @Override
+  public Future<Boolean> updateUploadDefinitionStatusIfAllFilesUploadingFailed(UploadDefinition uploadDefinition, UploadDefinition.Status status, String tenantId) {
```
KaterynaSenchenko: do we really need UploadDefinition.Status status argument for this method?

RuslanLavrov: at first the argument did not exist, but the second comment from @OleksiiKuzminov says that it's bad practice

OleksiiKuzminov: Can you just make method update status and call it if condition is true?

src/main/java/org/folio/service/upload/UploadDefinitionServiceImpl.java
```diff
@@ -378,6 +378,22 @@ private Errors validateFreeSpace(UploadDefinition definition, Errors errors) {
...
+    Future<Boolean> result = Future.succeededFuture(false);
+
+//    if(areAllDefinitionsFailed(uploadDefinition)) {
```
OleksiiKuzminov: ?

src/main/java/org/folio/service/upload/UploadDefinitionServiceImpl.java
```diff
@@ -378,6 +378,22 @@ private Errors validateFreeSpace(UploadDefinition definition, Errors errors) {
     }, tenantId);
   }
 
+  @Override
+  public Future<Boolean> updateUploadDefinitionStatusIfAllFilesUploadingFailed(UploadDefinition uploadDefinition, UploadDefinition.Status status, String tenantId) {
```
OleksiiKuzminov: we dont need this method

PR #72, opened 2019-05-07
MODDATAIMP-123 - Implement a blocking queue for posting chunks of raw records

Igor-Gorchakov:


src/main/java/org/folio/service/processing/coordinator/QueuedBlockingCoordinator.java
```diff
@@ -0,0 +1,45 @@
...
+    blockingQueue = new ArrayBlockingQueue<>(capacity, true);
+  }
+
+  @Override
+  public void acceptLock() {
+    try {
+      blockingQueue.put(QUEUE_ITEM);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      LOGGER.info("Failed to accept lock. The current thread " + Thread.currentThread().getName() + " is interrupted. Cause:" + e.getCause());
```
KaterynaSenchenko: please use log parameters instead of string concatenation

src/main/java/org/folio/service/processing/coordinator/QueuedBlockingCoordinator.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Override
+  public void acceptUnlock() {
+    try {
+      if (!blockingQueue.isEmpty()) {
+        blockingQueue.take();
+      }
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      LOGGER.info("Failed to accept unlock. The current thread " + Thread.currentThread().getName() + " is interrupted. Cause:" + e.getCause());
```
KaterynaSenchenko: same here

PR #77, opened 2019-05-16
MODSOURMAN-112 Implement a more elaborate way to define a RecordParser

Anatolii-Starkov:
-added contentType field to the RawRecordsDto


Anatolii-Starkov:
@Anatolii-Starkov I am waiting for that but thanks for your attention.

src/main/java/org/folio/service/processing/reader/MarcXmlReader.java
```diff
@@ -10,6 +10,7 @@
 import org.dom4j.DocumentException;
 import org.dom4j.Element;
 import org.dom4j.io.SAXReader;
+import org.folio.rest.jaxrs.model.RawRecordsDto.ContentType;
 
 /**
  * It reads marc records from an xml file by specified size of chunk.
```
OleksiiKuzminov: Plz use xml reader from marc4j lib

PR #85, opened 2019-06-26
MODDATAIMP-138 - Added exception handling of response body mapping to entity

RuslanLavrov:


src/main/java/org/folio/service/upload/UploadDefinitionServiceImpl.java
```diff
@@ -347,13 +348,23 @@ private Errors validateFreeSpace(UploadDefinition definition, Errors errors) {
     return future;
   }
 
+  private <T> Future<T> mapBufferContentToEntity(Buffer buffer, Class<T> entityType) {
```
OleksiiKuzminov: javadoc

src/main/java/org/folio/service/upload/UploadDefinitionServiceImpl.java
```diff
@@ -347,13 +348,30 @@ private Errors validateFreeSpace(UploadDefinition definition, Errors errors) {
...
   }
 
+  /**
+   * Returns instantiated T entity object from a buffer content.
+   *
+   * @param buffer buffer
+   * @param entityType type of entity which will be created
+   * @return instantiated T entity object from a buffer content
+   */
+  private <T> Future<T> mapBufferContentToEntity(Buffer buffer, Class<T> entityType) {
```
OleksiiKuzminov: Lets move this method to util class

PR #94, opened 2019-09-23
MODDATAIMP-192 - Apply blocking coordination for file processing

Igor-Gorchakov:
User can import multiple files and start file processing. Data-import does reading/sending records for all the files in parallel, that may overflow record-manager and record-storage.
We should limit file processing to do not send all the files at once:
apply BlockingCoordinator to handle files with given limit in sequential manner.


OleksiiKuzminov:
plz make sonar happy and ok

Igor-Gorchakov:
> plz make sonar happy and ok

Done!

PR #95, opened 2019-09-30
MODDATAIMP-191

OleksiiKuzminov:
Create a Preprocessor for import file to fix encoding issues


Igor-Gorchakov:
Do we have same issue reading XML/JSON ?

PR #120, opened 2020-05-04
MODDATAIMP-300 - Update version for marc4j

RuslanLavrov:


OleksiiKuzminov:
@RuslanLavrov plz add close to streams

PR #123, opened 2020-06-12
Moddataimp 304 Lack of details in case of an error during file processing

OleksiiKuzminov:


pom.xml
```diff
@@ -165,7 +165,7 @@
     <dependency>
       <groupId>org.folio</groupId>
       <artifactId>mod-source-record-manager-client</artifactId>
-      <version>2.2.0-SNAPSHOT</version>
+      <version>2.2.0</version>
```
KaterynaSenchenko, line 168: ???

PR #168, opened 2021-03-30
MODDATAIMP-400: Resolve data-import catching error issues [BUGFIX]

VRohach:
## Purpose
Job with empty or invalid marc file stucks with invalid (NaN%) progress number. It can be removed.
The main goal - to fix it.

## Approach
If there are no records or invalid MARC-content (just simple text, for example), then data-import will be finished with Failed status. The job won`t be stuck.

Moreover, added try/catch to the highest level for logging different errors.

## Learning
More info: https://issues.folio.org/browse/MODDATAIMP-400

### DIFF
--- a/src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
+++ b/src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
@@ -1,14 +1,18 @@
 package org.folio.service.processing;
 
 import org.folio.okapi.common.GenericCompositeFuture;
+
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
 import io.vertx.core.json.JsonObject;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+
 import io.vertx.ext.web.handler.impl.HttpStatusException;
 import io.vertx.kafka.client.producer.KafkaProducer;
+
 import org.apache.commons.collections4.list.UnmodifiableList;
 import org.folio.HttpStatus;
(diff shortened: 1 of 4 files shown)


RuslanLavrov (approved):

src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
```diff
@@ -66,23 +70,27 @@ public ParallelFileChunkingProcessor(Vertx vertx, KafkaConfig kafkaConfig) {
...
+        .compose(fileStorageService -> {
+          processFiles(jobProfile, uploadDefinitionService, fileStorageService, uploadDefinition, params);
+          uploadDefinitionService.updateBlocking(
+            uploadDefinition.getId(),
+            definition -> succeededFuture(definition.withStatus(UploadDefinition.Status.COMPLETED)),
+            params.getTenantId());
+          return succeededFuture();
+        });
+    } catch (Exception e) {
+      LOGGER.error("Can`t process file. Cause: ", e.getMessage());
```
KaterynaSenchenko, line 92: do we have JobExecution entity created at this point? if yes - we need to update its status to failed

VRohach: @KaterynaSenchenko  it seems like we don`t have this property here. I changed this try/catch to the ".onFailure()" with logging.

 I want to mention, that if there are some exceptions in the places where we have JobExecution entity - it`ll log and update status at the level below (processFile-method) because I added there one more error catching in the previous tickets (which fixed some bugs when data-import can`t started).

OleksiiKuzminov (approved):

PR #173, opened 2021-05-10
MODDATAIMP-433 Store MARC Authority record

oleksandr-dekin:
## Purpose
Change record type MARC for more specific type, like MARC_BIB

## Approach
- updated data type from marc to marc_bib

## Learning
[MODDATAIMP-433](https://issues.folio.org/browse/MODDATAIMP-433)


src/main/resources/templates/db_scripts/defaultFileExtensions.sql
```diff
@@ -283,7 +283,7 @@ VALUES
       "extension":".xml",
       "description": "",
       "dataTypes":[
-        "MARC"
+        "MARC_BIB"
```
psmagin, line 286: Maybe there should be more dataTypes? MARC_AUTH and MARC_HOLDINGS?

psmagin: And I assume that this script is already part of envs, so you need to do migration.

psmagin (changes requested):

src/main/java/org/folio/service/processing/reader/SourceReaderBuilder.java
```diff
@@ -49,6 +49,6 @@ private static boolean isEdifact(String extension, JobProfileInfo jobProfile) {
   }
 
   private static boolean isMarc(JobProfileInfo jobProfile) {
-    return jobProfile != null && jobProfile.getDataType() == JobProfileInfo.DataType.MARC;
+    return jobProfile != null && jobProfile.getDataType() == JobProfileInfo.DataType.MARC_BIB;
```
psmagin, line 52: And also there I assume should be more datatypes.

psmagin (approved):

PR #183, opened 2021-07-29
MODDATAIMP-459 EDIFACT files with CAPS file extensions do not import

bisecomsergiy:
## Purpose
When an invoice file has an upper case file extension (INV or EDI) or mixed case (inV or eDi), there difficulties to import such kind of files.

## Approach
With purpose to import edifact files with upper, lower or mixed case extensions, applied insensitive case logic. 

## Learning
https://issues.folio.org/browse/MODDATAIMP-459


Aliaksandr-Fedasiuk (approved):

src/test/java/org/folio/service/processing/reader/SourceReaderBuilderTest.java
```diff
@@ -36,6 +42,7 @@ public void buildShouldReturnMarcJsonReader() {
     //when
     SourceReader reader = SourceReaderBuilder.build(new File(SOURCE_JSON_PATH), marcJobProfile);
     //then
+    Assert.assertNotNull(reader);
```
SerhiiNosko, line 45: Reader can not be null in any case, because it is always previously created

KaterynaSenchenko (approved):

src/test/java/org/folio/service/processing/reader/SourceReaderBuilderTest.java
bisecomsergiy: The tests are updated, thanks

SerhiiNosko (approved):

PR #193, opened 2021-09-20
MODDATAIMP-548 - Provide system properties to set chunk size for each marc record format

RuslanLavrov:
## Purpose
since records chunk size depends on record type/format (e.g. xml format has more redundant syntax) it makes sense to have the ability to set records chunk size for particular record type separately

## Approach
* delete zipping operation for data-import payload before send to Kafka
* provide system properties for SourceReaderBuilder to set records chunk size for each record type and marc format (raw, json, xml) correspondingly
  * "file.processing.marc.raw.buffer.chunk.size" property with 50 size value by default for marc raw format
  *  "file.processing.marc.json.buffer.chunk.size" property with 50 size value by default for marc json format
  *  "file.processing.marc.xml.buffer.chunk.size" property with 10 size value by default for marc xml format (since the xml format is more redundant the proposed default value has been selected to not exceed "max.request.size" producer parameter which is 1048576 bytes by default)
  *   "file.processing.edifact.buffer.chunk.size" property with 50 size value by default for edifact record type

## Learning
[MODDATAIMP-548](https://issues.folio.org/browse/MODDATAIMP-548)

### DIFF
--- a/src/main/java/org/folio/service/processing/reader/SourceReaderBuilder.java
+++ b/src/main/java/org/folio/service/processing/reader/SourceReaderBuilder.java
@@ -2,12 +2,10 @@ package org.folio.service.processing.reader;
 
 import org.apache.commons.io.FilenameUtils;
 import org.folio.rest.jaxrs.model.JobProfileInfo;
-import org.folio.rest.jaxrs.model.JobProfileInfo.DataType;
 
 import java.io.File;
 
 import static java.util.Optional.ofNullable;
-import static org.folio.rest.RestVerticle.MODULE_SPECIFIC_ARGS;
 import static org.folio.service.processing.reader.EdifactReader.EDIFACT_EDI_EXTENSION;
 import static org.folio.service.processing.reader.EdifactReader.EDIFACT_INV_EXTENSION;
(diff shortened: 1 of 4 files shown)


SerhiiNosko (approved):

Aliaksandr-Fedasiuk (approved):

bisecomsergiy (approved):

AivarIUsupov (approved):

src/main/java/org/folio/service/processing/reader/SourceReaderBuilder.java
```diff
@@ -31,14 +34,14 @@ public static SourceReader build(File file, JobProfileInfo jobProfile) {
 
     if (isMarc(jobProfile)) {
       if (JSON_EXTENSION.equals(extension)) {
```
VRohach, line 36: @RuslanLavrov  I think it will be better to use switch-case here for more readability and code efficiency. At least, for now. What do you think?

RuslanLavrov: merging for now as is to be in time for hosted envs build

PR #214, opened 2022-03-25
MODDATAIMP-472 allow any extension to be processed by EDIFACT reader

okolawole-ebsco:
## Purpose

- Allow other file extensions to be consumed by the edifact reader
- Handle exceptions when reading a raw "MARC" file.
- Emit a DI_ERROR event when exceptions occur in mod-data-import 

## Approach
mod-data-import does not seem to be designed to emit DI_ERROR events. These events tend to not have records in them, which caused an issue downstream in mod-source-record-manager(SRM). SRM has been updated to handle DI_ERROR events that don't have records in another PR.
Changes in EventHandlingUtil in mod-data-import was modeled after EventHandlingUtil in SRM.


okolawole-ebsco:
Please let me know if the approach is good. I will then add tests to shore things up!

src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
```diff
@@ -12,15 +13,11 @@
...
 import org.folio.kafka.KafkaTopicNameHelper;
 import org.folio.okapi.common.GenericCompositeFuture;
 import org.folio.rest.client.ChangeManagerClient;
-import org.folio.rest.jaxrs.model.FileDefinition;
-import org.folio.rest.jaxrs.model.JobExecutionDto;
-import org.folio.rest.jaxrs.model.JobProfileInfo;
-import org.folio.rest.jaxrs.model.ProcessFilesRqDto;
-import org.folio.rest.jaxrs.model.StatusDto;
-import org.folio.rest.jaxrs.model.UploadDefinition;
+import org.folio.rest.jaxrs.model.*;
```
Abdulkhakimov, line 20: I recommend to use single import for necessary classes for avoiding naming collision.

src/main/java/org/folio/service/util/EventHandlingUtil.java
```diff
@@ -1,14 +1,116 @@
...
+    String producerName = eventType + "_Producer";
+    KafkaProducer<String, String> producer =
+      KafkaProducer.createShared(vertx, producerName, kafkaConfig.getProducerProps());
+    producer.write(record, war -> {
+      producer.end(ear -> producer.close());
+      if (war.succeeded()) {
+        logSendingSucceeded(eventType, chunkId, recordId);
+        promise.complete(true);
+      } else {
+        Throwable cause = war.cause();
```
SerhiiNosko, line 72: We need to fail promise like this: promise.fail(cause);

src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
okolawole-ebsco: Updated. My IDE automatically makes this change. I will look into it.

src/main/java/org/folio/service/util/EventHandlingUtil.java
```diff
@@ -1,14 +1,116 @@
...
+        LOGGER.error("{} write error for event {}:", producerName, eventType, cause);
+      }
+    });
+    return promise.future();
+  }
+
+  private static void logSendingSucceeded(String eventType, String chunkId, String recordId) {
+    if (recordId == null) {
```
SerhiiNosko, line 80: Better to check using StringUtils.isBlank or add checking on empty string also

src/main/java/org/folio/service/processing/ParallelFileChunkingProcessor.java
Abdulkhakimov: Thanks

src/main/java/org/folio/service/util/EventHandlingUtil.java
okolawole-ebsco: Agreed. This was modeled after code in SRM codebase.

src/main/java/org/folio/service/processing/reader/SourceReaderBuilder.java
```diff
@@ -40,16 +38,15 @@ public static SourceReader build(File file, JobProfileInfo jobProfile) {
...
     }
 
     return ofNullable(sourceReader).orElseThrow(() -> new UnsupportedOperationException("Unsupported file format"));
   }
 
-  private static boolean isEdifact(String extension, JobProfileInfo jobProfile) {
-    return ((jobProfile.getDataType() == JobProfileInfo.DataType.EDIFACT)
-      && (EDIFACT_EDI_EXTENSION.equalsIgnoreCase(extension) || EDIFACT_INV_EXTENSION.equalsIgnoreCase(extension)));
+  private static boolean isEdifact(JobProfileInfo jobProfile) {
+    return jobProfile != null && (jobProfile.getDataType() == JobProfileInfo.DataType.EDIFACT);
```
Abdulkhakimov, line 49: You can use Objects.nonNull method. The static method is doing code more readable.

src/main/java/org/folio/service/util/EventHandlingUtil.java
SerhiiNosko: We could start thinking to moving common codebase for EventHandlingUtils to folio-kafka-wrapper to avoid code duplication, but  its not in scope of this PR of course

src/test/java/org/folio/service/processing/ParallelFileChunkingProcessorUnitTest.java
```diff
@@ -14,10 +14,7 @@
 import org.folio.kafka.KafkaConfig;
 import org.folio.kafka.KafkaTopicNameHelper;
 import org.folio.rest.AbstractRestTest;
-import org.folio.rest.jaxrs.model.Event;
-import org.folio.rest.jaxrs.model.FileDefinition;
-import org.folio.rest.jaxrs.model.JobProfileInfo;
-import org.folio.rest.jaxrs.model.RawRecordsDto;
+import org.folio.rest.jaxrs.model.*;
```
Abdulkhakimov, line 17: I forgot to mention this line. Please remove wildcard import from tests too.

src/main/java/org/folio/service/util/EventHandlingUtil.java
okolawole-ebsco: Agreed as well. I wanted to do that but decided against it so that my first PR is not bloated.

src/test/java/org/folio/service/processing/ParallelFileChunkingProcessorUnitTest.java
okolawole-ebsco: updated.

src/main/java/org/folio/service/processing/reader/SourceReaderBuilder.java
okolawole-ebsco: updated

src/main/java/org/folio/service/util/EventHandlingUtil.java
okolawole-ebsco: updated.

SerhiiNosko (approved):

Abdulkhakimov (approved):

PR #238, opened 2022-12-21
[MODDATAIMP-641] Logging improvement

RomanChernetskyi:
## Purpose
Logging improvement

## Approach
Added new and changed old logs

### DIFF
--- a/src/main/java/org/folio/dao/FileExtensionDaoImpl.java
+++ b/src/main/java/org/folio/dao/FileExtensionDaoImpl.java
@@ -61,7 +61,7 @@ public class FileExtensionDaoImpl implements FileExtensionDao {
       CQLWrapper cql = getCQLWrapper(FILE_EXTENSIONS_TABLE, query, limit, offset);
       pgClientFactory.createInstance(tenantId).get(FILE_EXTENSIONS_TABLE, FileExtension.class, fieldList, cql, true, false, promise);
     } catch (Exception e) {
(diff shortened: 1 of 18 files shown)


KaterynaSenchenko (approved):

src/main/java/org/folio/dao/FileExtensionDaoImpl.java
```diff
@@ -113,6 +113,7 @@ public Future<Optional<FileExtension>> getFileExtensionByExtension(String extens
 
   @Override
   public Future<String> addFileExtension(FileExtension fileExtension, String tenantId) {
+    LOGGER.trace("addFileExtension:: adding file extension {} for tenant {}", fileExtension.getId(), tenantId);
```
Aliaksandr-Fedasiuk, line 116: maybe debug level will be better?

Aliaksandr-Fedasiuk (commented):
@RomanChernetskyi, could you add logs for this recommendation:  
Controller endpoints or methods containing business must have at least two messages:
    method enter: containing method name and incoming parameters, DEBUG level

src/main/java/org/folio/dao/FileExtensionDaoImpl.java
RomanChernetskyi: refactored DAO logs

KyryloShulzhenko (approved):

KaterynaSenchenko (approved):

PR #253, opened 2023-06-13
MODDATAIMP-846: Add GET /uploadUrl endpoint

ncovercash:
## Purpose
As part of the data import improvement project, we want to upload directly to S3-like storage from the frontend.  This requires obtaining a presigned URL and exposing an endpoint to obtain this URL, which was added as `GET /uploadUrl` (see https://github.com/folio-org/folio-s3-client/pull/11 for the accompanying changes in `folio-s3-client`).

## Approach
This adds a new endpoint `GET /uploadUrl` and entity `FileUploadInfo` to contain data that allows the frontend client to upload directly to S3.  To accomplish this, infrastructure was added to interface with the FOLIO S3 client.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,19 @@ To allow multiple instance deployment, for every instance the same persistent vo
 * **_data.import.storage.type_** - type of data storage used for uploaded files. Default value is **LOCAL_STORAGE**. Other implementations for storage should be added.
 * **_data.import.storage.path_** - **path where uploaded file will be stored**
 
+## Interaction with AWS S3/Minio
+
+This module uses S3-compatible storage as part of the file upload process.  The following environment variables must be set with values for your S3-compatible storage (AWS S3, Minio Server):
+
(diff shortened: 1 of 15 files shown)


ncovercash:
I'm not sure why Jenkins is failing...it looks like all the tests pass, but Docker is having some kind of issue.  Can anyone provide any insight into this?

carolegodfrey (approved):
Thanks for adding tests (including failures)  @ncovercash -- I tested this branch against Rancher and was able to generate a presigned URL for PUT and issued a PUT request against this URL to upload a file to S3

ncovercash:
> Thanks for adding tests (including failures) @ncovercash -- I tested this branch against Rancher and was able to generate a presigned URL for PUT and issued a PUT request against this URL to upload a file to S3

As we saw yesterday, we can't assume AWS will always work 😉

Aliaksandr-Fedasiuk (approved):

PR #255, opened 2023-06-26
Update mocking style for DAO tests

ncovercash:


src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -37,35 +41,58 @@ public class DataImportQueueDaoTest  {
 
   @InjectMocks
   DataImportQueueItemDao queueItemDaoImpl = new DataImportQueueItemDaoImpl();
-  
 
   UUID storedItemUUID;
+
   @Before
   public void setUp() {
-    MockitoAnnotations.initMocks(this);
```
ncovercash, line 45: `initMocks` is deprecated

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -37,35 +41,54 @@ public class DataImportQueueDaoTest  {
...
     when(postgresClientFactory.getInstance()).thenReturn(postgresClient);
   }
-  @Test 
+
+  @Test
   public void shouldAddQueueItem(TestContext context) {
     // given
-    when(postgresClient.execute(anyString(), any(Tuple.class)))
-      .thenReturn(Future.succeededFuture(new LocalRowSet(1)));
+    doAnswer((InvocationOnMock invocation) -> {
```
ncovercash, line 56: This was the main change.  When mocking `void` functions, the normal `when(mock.method(parameters)).thenAnswer/thenReturn/etc` does not work; instead, it has to be flipped to be `doAnswer(...).when(mock).method(parameters)` due to some technicalities with the way `void` works.

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -37,35 +41,54 @@ public class DataImportQueueDaoTest  {
...
+        Promise<RowSet<Row>> promise = invocation.getArgument(2);
```
ncovercash, line 57: Have to do some funkiness here to cast the argument to a `Promise`, since `Promise` is parameterized.  Normally, we could just do `invocation.getArgument(2, Promise.class).complete(...)` but for this we need that extra variable

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -37,35 +41,54 @@ public class DataImportQueueDaoTest  {
...
+        promise.complete(new LocalRowSet(1));
```
ncovercash, line 58: The reason your tests were hanging is that you were returning a result from `.execute`, but the test was waiting on the promise/future to be completed.

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -37,35 +41,54 @@ public class DataImportQueueDaoTest  {
...
+        return null;
+      })
+      .when(postgresClient)
+      .execute(
+        anyString(),
+        any(Tuple.class),
+        ArgumentMatchers.<Promise<RowSet<Row>>>any()
```
ncovercash, line 65: It looks really ugly, but this is the way to make an `any()` that accepts a generic (since we can't do `.class` on a generic, otherwise we'd be able to just use the syntax like `Tuple.class` uses.

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -37,35 +41,54 @@ public class DataImportQueueDaoTest  {
...
+      // then
+      .onComplete(
+        context.asyncAssertSuccess(x -> {
+          verify(postgresClient, times(1))
+            .execute(
+              anyString(),
+              any(Tuple.class),
+              ArgumentMatchers.<Promise<RowSet<Row>>>any()
+            );
+          verifyNoMoreInteractions(postgresClient);
```
ncovercash, line 90: Always good to include this in your test; it will ensure mocked calls are made and that no other calls were made.  I think this might have also caught the mocking error (it will tell you the exact issue, e.g.):
```

Argument(s) are different! Wanted:
postgresClient.execute(
    <any string>,
    <any io.vertx.sqlclient.Tuple>
);
-> at org.folio.rest.persist.PostgresClient.execute(PostgresClient.java:3869)
Actual invocations have different arguments:
postgresClient.execute(
    "INSERT INTO data_import_global.queue_items (id, jobExecutionId,uploadDefinitionId,size,originalSize,filePath,timestamp) VALUES ($1, $2, $3, $4, $5, $6, $7)",
    io.vertx.sqlclient.impl.ArrayTuple@40de8f93,
    Future{result=org.folio.rest.persist.helpers.LocalRowSet@6ff0b1cc}
);
```

evaluk (approved):

PR #256, opened 2023-06-27
Test fixes

ncovercash:


src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -54,118 +66,170 @@ public void setUp() {
   public void shouldGetQueueItem(TestContext context) {
     // given
     doAnswer((InvocationOnMock invocation) -> {
-      Promise<RowSet<Row>> promise = invocation.getArgument(2);
-      promise.complete(new LocalRowSet(5));
-      return null;
-    }).when(postgresClient)
-      .execute(anyString(), any(Tuple.class), ArgumentMatchers.<Promise<RowSet<Row>>>any());
+        Promise<RowSet<Row>> promise = invocation.getArgument(1);
```
ncovercash, line 69: Two issues here:
- you were mocking `.execute` when the call was to `.select`
- `.select` has the `Promise` as the second argument, so we need `.getArgument(1)

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -54,118 +66,170 @@ public void setUp() {
...
       // then
-      .onComplete(context.asyncAssertSuccess(x -> {
-        verify(postgresClient, times(1)).execute(anyString(), any(Tuple.class), ArgumentMatchers.<Promise<RowSet<Row>>>any());
-        verifyNoMoreInteractions(postgresClient);
-      }));
+      .onComplete(
+        context.asyncAssertSuccess(x -> {
+          verify(postgresClient, times(1))
+            .select(
+              startsWith("SELECT * FROM data_import_global.queue_items"),
```
ncovercash, line 85: For extra confidence, you can use matchers like `startsWith` or `contains` to ensure the "run" SQL queries look the way they should

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -54,118 +66,170 @@ public void setUp() {
...
       // then
-      .onComplete(context.asyncAssertSuccess(x -> {
-        verify(postgresClient, times(1)).execute(anyString(), any(Tuple.class), ArgumentMatchers.<Promise<RowSet<Row>>>any());
-        verifyNoMoreInteractions(postgresClient);
-      }));
+      .onComplete(
+        context.asyncAssertSuccess(x -> {
+          verify(postgresClient, times(1))
+            .execute(
+              startsWith("INSERT INTO data_import_global.queue_items "),
```
ncovercash, line 126: All that changed here was adding a more specific matcher

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -54,118 +66,170 @@ public void setUp() {
...
   }
 
   @Test
-  public void shouldGetQueueItemById(TestContext context) {
+  public void shouldGetQueueItemByIdFailure(TestContext context) {
+    Async async = context.async();
+
     // given
     doAnswer((InvocationOnMock invocation) -> {
-      Promise<RowSet<Row>> promise = invocation.getArgument(2);
```
ncovercash, line 108: Mocking here looked good 👍

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -54,118 +66,170 @@ public void setUp() {
...
+        any(Tuple.class),
+        ArgumentMatchers.<Promise<RowSet<Row>>>any()
+      );
 
     // when
-    queueItemDaoImpl.getQueueItemById("sdfasdfasdf")
+    queueItemDaoImpl
+      .getQueueItemById("sample-id")
       // then
-      .onComplete(context.asyncAssertSuccess(x -> {
```
ncovercash, line 117: Since we expect this to fail, I used `.onFailure` instead.  This doesn't return an `AsyncResult` though; instead, it returns the exception it failed with.

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -54,118 +66,170 @@ public void setUp() {
...
+        verify(postgresClient, times(1))
+          .select(
+            anyString(),
+            any(Tuple.class),
+            ArgumentMatchers.<Promise<RowSet<Row>>>any()
+          );
         verifyNoMoreInteractions(postgresClient);
-      }));
+
+        assertThat(err, is(instanceOf(NoSuchElementException.class)));
```
ncovercash, line 165: To test the exception returned, I used this matcher here.

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -54,118 +66,170 @@ public void setUp() {
...
+
+        async.complete();
```
ncovercash, line 167: Since we're not getting an `AsyncResult`, the `context.assertAsyncSuccess` will not work.  However, we can manually control success/failure by running `context.async()` at the top of the test then running `async.success()` and `async.fail()` when the test succeeds/fails

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -54,118 +66,170 @@ public void setUp() {
...
+      })
+      .onSuccess(er -> context.fail("Provided ID should not exist."));
   }
 
   @Test
   public void shouldUpdateQueueItemById(TestContext context) {
     // given
     doAnswer((InvocationOnMock invocation) -> {
-      Promise<RowSet<Row>> promise = invocation.getArgument(2);
```
ncovercash, line 127: More great mocking!

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -54,118 +66,170 @@ public void setUp() {
...
+            );
+          verifyNoMoreInteractions(postgresClient);
+        })
+      );
   }
+
   @Test
   public void shouldDeleteQueueItemById(TestContext context) {
     // given
-    doAnswer((InvocationOnMock invocation) -> {
```
ncovercash, line 156: This is a bit backwards to the others; for this one, the `postgresClient.execute` **does** return a value, so we'll use the traditional `when(object.method(...)).thenReturn(...)`.  We only have to do it the other way around (with `do`) when the function is void

ncovercash (commented):
Lmk if you have any questions!  Overall, looks great, and you're absolutely on the right track.

If all looks good/clear to you, feel free to merge at-will

evaluk (approved):

PR #257, opened 2023-06-29
[DRAFT] [MODDATAIMP-864] Create a service to implement task force's prioritization algorithm

ncovercash:
This is a draft to be able to run my changes through CI and increase visibility into my work.

_I'll change the base to Orchid once Eric's work is merged; I don't want to cause any weird merge/history weirdness._

### DIFF
--- a/src/main/java/org/folio/service/processing/ranking/ScoreService.java
+++ b/src/main/java/org/folio/service/processing/ranking/ScoreService.java
@@ -1,0 +1,10 @@
+package org.folio.service.processing.ranking;
+
+import org.folio.rest.jaxrs.model.DataImportQueueItem;
+
+public interface ScoreService {
+  /**
+   * Calculate the score for a given chunk based on the class's metric/implementation
+   */
+  double calculateScore(DataImportQueueItem chunk);
+}
--- a/pom.xml
+++ b/pom.xml
@@ -258,6 +258,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-all</artifactId>
+      <version>1.3</version>
+    </dependency>
+    <dependency>
       <groupId>net.mguenther.kafka</groupId>
       <artifactId>kafka-junit</artifactId>
       <version>${kafka-junit.version}</version>
--- a/src/main/java/org/folio/service/processing/ranking/ScoreUtils.java
+++ b/src/main/java/org/folio/service/processing/ranking/ScoreUtils.java
@@ -1,0 +1,133 @@
+package org.folio.service.processing.ranking;
+
+import javax.validation.constraints.Min;
+
+public class ScoreUtils {
+
+  private ScoreUtils() {
+    throw new UnsupportedOperationException(
+      "Cannot instantiate utility class."
+    );
+  }
+
+  /**
+   * Calculate the logarithm-based score for a given value [0,inf).
+   *
+   * For example, if we expect {@code value} to be from 0-31 (reference=31),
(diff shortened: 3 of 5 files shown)


src/main/java/org/folio/service/processing/ranking/ScoreService.java
```diff
@@ -0,0 +1,140 @@
...
+    return CompositeFuture
+      .all(
+        queueItemDao.getAllInProgressQueueItems(),
+        queueItemDao.getAllWaitingQueueItems()
+      )
+      .map((CompositeFuture result) -> {
+        DataImportQueueItemCollection inProgress = result.resultAt(0);
+        DataImportQueueItemCollection waiting = result.resultAt(1);
+
+        Map<String, Long> tenantUsageMap = getTenantUsageMap(inProgress);
```
evaluk, line 63: the tenant usage again seems to hit the database, this may be a price we can pay but there are in memory options if this turns out to be expensive.

src/main/java/org/folio/service/processing/ranking/ScoreService.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+        return Optional.of(set.first());
+      });
+  }
+
+  /**
+   * Get a list of all WAITING queue items, sorted by score, with the highest
+   * score coming first
+   */
+  public Future<NavigableSet<DataImportQueueItem>> getRankedQueueItems() {
```
evaluk, line 47: I was trying to think through how this function would work with multiple threads calling it.  Do we have multiple workers calling it?  Does the process of sending a file off to data import require parallel execution?  This would preclude the need for tracking in progress items to load balance tenants(unless we loop back around to tracking items in actual progress with further status). 

I guess this was a long winded way to say we might not need the usage scoring.

src/main/java/org/folio/service/processing/ranking/ScoreService.java
```diff
@@ -0,0 +1,140 @@
...
+    // than one request/sorting.
+    Map<DataImportQueueItem, Double> cache = new HashMap<>();
+
+    return CompositeFuture
+      .all(
+        queueItemDao.getAllInProgressQueueItems(),
+        queueItemDao.getAllWaitingQueueItems()
+      )
+      .map((CompositeFuture result) -> {
+        DataImportQueueItemCollection inProgress = result.resultAt(0);
```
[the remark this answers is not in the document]
ncovercash: I did, but since we're going to have multiple workers, we'd need to perform queries anyways to determine what jobs have been taken since, which tenants are using most resources, etc.; two `SELECT` calls is a very small performance hit IMO.

src/main/java/org/folio/service/processing/ranking/ScoreService.java
ncovercash: This uses the same results from the DB calls; only two queries are needed for the entire process.

src/main/java/org/folio/service/processing/ranking/ScoreService.java
ncovercash: I'm not sure of the best way to make this concurrent; perhaps I should add some functionality to "claim" a job (mark it as in-progress) and make this entire process atomic, such that only one worker could do it at a time.

ncovercash: I'll work on that tomorrow; I also planned on adding some more test cases

evaluk (approved):
Looks good, hopefully it will play out in performance testing whether this hits the DB too much

KaterynaSenchenko (approved):

PR #258, opened 2023-06-29
MODDATAIMP-830: create public schemas

evaluk:
# [Jira MODDATAIMP-830](https://issues.folio.org/browse/MODDATAIMP-830)

## Purpose
This pull request adds a new schema to store split file information for use later by queueing mechanisms.

## Approach
This story creates a liquibase table and a data access object to use with the new table.  Unit tests are also implemented in mockito to demonstrate the functionality.

Changes also located in [MODDATAIMP_830](https://github.com/folio-org/data-import-raml-storage/pull/287) for raml
#### TODOS and Open Questions

Possible need to add more columns to the table.  Will need follow on user stories.

## Learning
This implementation was modelled heavily on mod-pubsub since it deals in global table space and accesses them the same way.  The DAO and unit tests are basically the same template.


ncovercash (approved):
LGTM, just fix those Sonar smells (should be straightforward)

Aliaksandr-Fedasiuk (approved):

PR #259, opened 2023-07-07
[MODDATAIMP-868] Add part #, tenant, and progress flag to queue

ncovercash:
# [Jira MODDATAIMP-868](https://issues.folio.org/browse/MODDATAIMP-868)

- Should be merged with https://github.com/folio-org/data-import-raml-storage/pull/289

## Purpose
This adds the following fields/columns to queue items:
- **part number:** parts within a job will be delivered from the queue in order (rather than a nondeterministic random mess).
- **tenant:** we can tell which tenant a chunk belongs to so that limiting/weighting can happen accordingly
- **progress:** we can tell which chunks have not yet been started, while also keeping track of which ones are currently running so that no tenant saturates the worker pool

## Approach
This builds on #258, adding new fields and tests to the main classes created there.


carolegodfrey (approved):
LGTM!

RomanChernetskyi (approved):

src/main/resources/liquibase/module/scripts/v-0.0.1/2023-07-07--00-20-add-queue-item-columns.xml
```diff
@@ -0,0 +1,12 @@
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+  <changeSet id="2023-07-07--00-20-add-queue-item-columns" author="ncovercash">
```
RuslanLavrov, line 5: does it make sense to add these columns in the scope of the previously added script for the "dataImportQueueItems" table creation since that script has not been released so the table structure can be fully described there, what do you think?

RuslanLavrov (approved):
please, don't forget to apply "squash and merge" strategy during the PR merge so that we could keep Orchid release branch or master branch clear, in this way, it will be a 1 commit in the target branch after the merge. Thank you!!
image

src/test/java/org/folio/dao/DataImportQueueDaoTest.java
```diff
@@ -168,6 +164,80 @@ public void shouldGetQueueItemByIdFailure(TestContext context) {
...
+    // when
+    queueItemDaoImpl
+      .getAllWaitingQueueItems()
+      // then
+      .onComplete(
+        context.asyncAssertSuccess(x -> {
+          verify(postgresClient, times(1))
+            .select(
+              startsWith(
+                "SELECT * FROM data_import_global.queue_items WHERE processing = "
```
evaluk, line 194: should this be extended to check if the boolean actually equals true or false?  otherwise there is no difference between this test case and the one after it?

PR #260, opened 2023-07-11
Moddataimp 852 configure api for workflow

evaluk:
## Purpose
To provide an api endpoint that returns configuration status for splitting of data import files to the UI for use in its internal flow and messaging.  Defaults to false if no value is present in the classpath.

## Approach
Creates a spring @value variable which will pull the status variable from environment variables or properties.  

#### TODOS and Open Questions
I considered adding an affirmative test where it set the environment variable and then returned it, however this is simply testing spring functionality for injecting that value.

### DIFF
--- a/ramls/dataImport.raml
+++ b/ramls/dataImport.raml
@@ -32,7 +32,7 @@ types:
   dataImportEventTypes: !include raml-storage/schemas/common/dataImportEventTypes.json
   dataImportInitConfig: !include raml-storage/schemas/common/dataImportInitConfig.json
   fileUploadInfo: !include fileUploadInfo.json
-
+  splitStatus: !include raml-storage/schemas/dto/splitStatus.json
 
 traits:
   validate: !include raml-storage/raml-util/traits/validation.raml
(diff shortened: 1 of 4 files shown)


KaterynaSenchenko (approved):

ramls/dataImport.raml
```diff
@@ -309,3 +309,23 @@ resourceTypes:
           body:
             text/plain:
               example: "Internal server error"
+  /splitStatus:
```
RuslanLavrov, line 312: please, add the newly defined endpoint to the ModuleDescriptor-template.json file so that it will be available

RuslanLavrov: Eric, propose to make the endpoint path in hyphen-separated (split-status) format instead of camel case format just ensure compliance with project conventions https://dev.folio.org/guidelines/naming-conventions/#:~:text=REST%20JSON%20interfaces.-,API%20endpoints,-The%20back%2Dend
Thanks!

RuslanLavrov (approved):

ramls/dataImport.raml
evaluk: I added an entry to the descriptor and gave it a permission for access.

PR #261, opened 2023-07-13
[DRAFT] MODDATAIMP-860 - Create Service to Split Marc files

carolegodfrey:
## Purpose
https://issues.folio.org/browse/MODDATAIMP-860 - adds service methods for 

- Counting number of records in a Marc File
- Splitting a Marc file info multiple files

- Additionally adds method to read a file from S3 Storage and to Write a file to S3 Storage
- Added property - `forcePathStyle`  - `true` setting was needed when testing against Rancher Minio - typically setting is `false` for S3

## Approach
Uses presence of Marc Record Terminator (RT) to count records and to split records into multiple files (based on record count). Does not use Marc4J library - reads/writes buffered byte buffer 

## Notes
On security hotspot - this was present in s3 library as well
https://github.com/folio-org/folio-s3-client/pull/8
https://docs.sonarsource.com/sonarqube/9.7/user-guide/security-hotspots/ 

#### TODOS and Open Questions
- [ ] Need to improve code for `splitFile` method (Add a way to inject S3StorageWriter or write of output files locally)
- [ ] Need to add tests which validate split file content is as expected - code improvement - write files locally for test 
- [ ] Need to add tests for added S3 functionality (S3StorageWriter, MinioStorageServiceImpl new methods)
- [ ] Was unable to extend https://github.com/folio-org/folio-s3-client/blob/master/src/main/java/org/folio/s3/client/RemoteStorageWriter.java class - have a local class 
https://github.com/folio-org/mod-data-import/blob/MODDATAIMP-860-carole/src/main/java/org/folio/service/s3storage/S3StorageWriter.java main BufferedOutputStream vs BufferedWriter

## Learning
(https://www.loc.gov/marc/specifications/specrecstruc.html)
https://github.com/marc4j/marc4j
https://github.com/xbib/marc


carolegodfrey:
Refinement as to how this will be done

PR #262, opened 2023-07-16
[DRAFT] MODDATAIMP-860 - Create Service to Split Marc files - rev2

carolegodfrey:
## Purpose
Revised version for 
https://github.com/folio-org/mod-data-import/pull/261

Exploring as alternative to initial approach

## Approach
Pre-plans the file splitting process

- Read file stream a first time to get all record terminator positions

- Use this info to get count of number of records and starting, ending byte position for each record  in the file that is to be split

- Based on this - will have a plan for splitting of files

- Read file stream a second time and perform the split

- For each file to be written - will have starting and ending position of beginning and ending record needed

#### TODOS and Open Questions
- [ ] Exploring as alternative to first implementation
- [ ] Issue with the rewind of the stream


carolegodfrey:
Refinement to how this is done

PR #266, opened 2023-07-24
Moddataimp 874 implement api to combine s3 multi part uploads

evaluk:
## Purpose
Create an API that allows UI to complete an S3 upload.  This api takes in a list of tags created during upload, the file id created by a previous api and the file name called a key.

## Approach
It uses the Folio S3 client functionality built earlier to complete an upload.

#### TODOS and Open Questions
None other then further unit test possibly needed.

## Learning
eTags are retrieved by the UI as a consequence of uploading file parts to the presigned urls provided.  upload ID is not related to folio ids

### DIFF
--- a/src/main/java/org/folio/rest/impl/DataImportImpl.java
+++ b/src/main/java/org/folio/rest/impl/DataImportImpl.java
@@ -501,7 +501,7 @@ public class DataImportImpl implements DataImport {
           entity.getKey()
         );
         minioStorageService.completeMultipartFileUpload(entity.getKey(),  entity.getUploadId(),entity.getTags())
-          .map(  completed -> completed ? PostDataImportAssembleStorageFileResponse.respond204() : PostDataImportAssembleStorageFileResponse.respond400WithTextPlain("Failed to assemble Data Import upload file") )
+          .map(  completed -> Boolean.TRUE.equals(completed) ? PostDataImportAssembleStorageFileResponse.respond204() : PostDataImportAssembleStorageFileResponse.respond400WithTextPlain("Failed to assemble Data Import upload file") )
           .map(Response.class::cast)
           .otherwise(ExceptionHelper::mapExceptionToResponse)
           .onComplete(asyncResultHandler);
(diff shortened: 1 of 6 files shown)


ncovercash (changes requested):
fix those smells and it's all good from me!

evaluk:
> fix those smells and it's all good from me!
I fixed one of them because it was my mistake in copying the case but the boolean smell is something I do not think I can fix.  The boolean is being passed in to the future as an object.

src/main/java/org/folio/rest/impl/DataImportImpl.java
```diff
@@ -489,6 +490,27 @@ public void getDataImportSplitStatus(Map<String, String> okapiHeaders, Handler<A
...
+  public void postDataImportAssembleStorageFile(AssembleFileDto entity, Map<String, String> okapiHeaders,
+      Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    vertxContext.runOnContext(v -> {
+      try {
+        LOGGER.debug(
+          "postDataImportAssembleStorageFile:: Assemble Storage File to complete upload {}",
+          entity.getKey()
+        );
+        minioStorageService.completeMultipartFileUpload(entity.getKey(),  entity.getUploadId(),entity.getTags())
+          .map(  completed -> completed ? PostDataImportAssembleStorageFileResponse.respond204() : PostDataImportAssembleStorageFileResponse.respond400WithTextPlain("Failed to assemble Data Import upload file") )
```
ncovercash, line 504: Here's the solution to that smell:

```suggestion
          .map(  completed -> Boolean.TRUE.equals(completed) ? PostDataImportAssembleStorageFileResponse.respond204() : PostDataImportAssembleStorageFileResponse.respond400WithTextPlain("Failed to assemble Data Import upload file") )
```

This happens since `completed` is an object `Boolean`, rather than the primitive; therefore, it can theoretically be `null`.  Additionally, objects can't normally be tested using traditional `==`/`!=`/etc, so we want to use `.equals`

ncovercash (changes requested):

ncovercash (approved):
LGTM!

PR #271, opened 2023-08-14
[MODDATAIMP-894] Remove unused DI queue column `size`

ncovercash:
# [Jira MODDATAIMP-894](https://issues.folio.org/browse/MODDATAIMP-894)

## Purpose

We opted not to consider chunk size as a metric for prioritizing queue items; therefore, this column is no longer needed.

## Approach
This removes the support from the DAO and provides a Liquibase changelog to drop the column in the database.

### DIFF
--- a/src/main/java/org/folio/dao/DataImportQueueItemDao.java
+++ b/src/main/java/org/folio/dao/DataImportQueueItemDao.java
@@ -58,7 +58,7 @@ public interface DataImportQueueItemDao {
    * Saves {@link DataImportQueueItem} to database
    *
    * @param DataImportQueueItem DataImportQueueItem to save
-   * @return future
+   * @return future with added row's ID
    */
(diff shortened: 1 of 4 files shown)


ramls/raml-storage
evaluk, line 1: is this supposed to be here?  asking partially for my own information.  I often see submodule changes in my git but I do not commit them because I dont think I should.

evaluk (approved):
approved if the submodule change is intended

ramls/raml-storage
ncovercash: @evaluk i updated it to sync it with the raml-storage changes (PR https://github.com/folio-org/data-import-raml-storage/pull/294); imo it's good to commit when you make changes in raml-storage that are needed, since that way the generated files are consistent (and any new code that relies on them can pass CI/etc)

PR #272, opened 2023-08-15
[MODDATAIMP-888] Add functionality to register split files with SRM and queue

ncovercash:
# [Jira MODDATAIMP-888](https://issues.folio.org/browse/MODDATAIMP-888)

## Purpose
This adds features to register split files with mod-srm (as JobExecutions) and insert their information into the queue.

## Approach
This PR creates a service which can be called to perform all necessary steps.

## Notes
Some changes from MODDATAIMP-894 got squash-merged previously, but this branch is based on that one, so some chages from it may appear in this PR.

#### TODOS and Open Questions
- [x] Additional unit tests


src/main/java/org/folio/service/file/SplitFileProcessingService.java
```diff
@@ -0,0 +1,133 @@
...
+          } catch (Exception e) {
+            promise.fail(e);
+          }
+        }
+      );
+      partNumber++;
+    }
+
+    return CompositeFuture.join(
+      futures.stream().map(Future.class::cast).collect(Collectors.toList())
```
okolawole-ebsco, line 130: just for my information, why was the cast necessary?

ncovercash: The version of vert.x that we're on has some funky behavior with `CompositeFuture` where `CompositeFuture` only takes untyped `Future`s (or in other words, every future used in it must be a plain `Future`, not `Future<Anything>`, and also when we retrieve the values they'll need casting too (otherwise they're stuck as plain `Object`).

evaluk (approved):
seems fine to me

KaterynaSenchenko (approved):

src/main/java/org/folio/service/file/SplitFileProcessingService.java
```diff
@@ -0,0 +1,133 @@
...
+    this.queueItemDao = queueItemDao;
+  }
+
+  /**
+   * Registers split parts as Job Executions in mod-source-record-manager
+   * and adds each part to the DI queue.
+   *
+   * @return a {@link CompositeFuture} of {@link JobExecutionDto}
+   */
+  public CompositeFuture registerSplitFiles(
```
RuslanLavrov, line 51: could you please add a note to the comment for the method parameters to describe what `keys` means. Does it mean a path/location of the part of the original split file?

src/main/java/org/folio/service/file/SplitFileProcessingService.java
```diff
@@ -0,0 +1,133 @@
...
+              );
+              throw new HttpException(
+                response.result().statusCode(),
+                "Error creating new JobExecution"
+              );
+            } else {
+              BufferMapper
+                .mapBufferContentToEntity(
+                  response.result().bodyAsBuffer(),
+                  JobExecutionDtoCollection.class
```
RuslanLavrov, line 101:
```suggestion
                  InitJobExecutionsRsDto.class
```
I've found that this endpoint seems to return a response that corresponds to another json schema https://github.com/folio-org/mod-source-record-manager/blob/9dc47fd905ce2f90bd7368dc15e26d7180a1287c/ramls/change-manager.raml#L67 and we should use InitJobExecutionsRsDto class for response deserialization to avoid error during deserialization

src/main/resources/liquibase/module/scripts/v-0.0.1/2023-08-14--00-00-remove-queue-item-size-column.xml
```diff
@@ -0,0 +1,8 @@
+<databaseChangeLog
+  xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
+  <changeSet id="2023-08-14--00-00-remove-queue-item-size-column" author="ncovercash">
+    <dropColumn tableName="dataImportQueueItems" columnName="size" />
```
RuslanLavrov, line 6: would like to propose deleting this column in the initial script where the table is created instead of adding an extra script since this table has not been included in any module release yet and therefore this table is still under development

src/main/java/org/folio/service/file/SplitFileProcessingService.java
ncovercash: Yes; key is basically a file name for S3

src/main/resources/liquibase/module/scripts/v-0.0.1/2023-08-14--00-00-remove-queue-item-size-column.xml
ncovercash: I'd prefer to keep it together, at least until a later point in DITF's work; otherwise, we'd have a lot of issues with all our local dev environments + Rancher

src/main/java/org/folio/service/file/SplitFileProcessingService.java
ncovercash: good catch, thanks!

src/main/java/org/folio/service/file/SplitFileProcessingService.java
RuslanLavrov: got it thanks! Propose to add a note in the method's comment with a brief description of the method's parameters using @param Javadoc tag.

RuslanLavrov (approved):

src/main/java/org/folio/service/file/SplitFileProcessingService.java
RuslanLavrov: Noah, there is a dedicated util class GenericCompositeFuture to create CompositeFuture by passing a list of parameterized `Future<Anything>` if you prefer we can use it here.

PR #273, opened 2023-08-18
[MODDATAIMP-842] Add endpoint to get presigned download URLs

ncovercash:
## Purpose
This PR adds an endpoint which can be used by the UI at any time (during or after completion) to get a presigned URL that allows downloading a split file chunk from S3.

## Approach
- adds a suitable method to the minio client
- exposes the existing `getJobExecutionById` method in the `UploadDefinitionService`

## ✨ Bonus ✨ 
- refactored existing minio tests to use better vert.x-style assertions


evaluk (approved):

KaterynaSenchenko (approved):

ramls/dataImport.raml
```diff
@@ -176,6 +177,15 @@ resourceTypes:
               body:
                 text/plain:
                   example: "Internal server error"
+  /jobExecutions/{jobExecutionId}:
+    /downloadUrl:
```
RuslanLavrov, line 181: please, add the newly implemented endpoint to the moduleDescriptorTemplate.json file

descriptors/ModuleDescriptor-template.json
```diff
@@ -296,6 +296,11 @@
           "pathPattern": "/data-import/uploadUrl/subsequent",
           "permissionsRequired": ["data-import.uploadUrl.get"]
         },
+        {
+          "methods": ["GET"],
```
RuslanLavrov, line 300: Could you please update the 'methods' property and the permission (as data-import.downloadUrl.post) with POST  to reflect the method of this endpoint? Also pls remove the redundant '/' in the pathPattern.

ncovercash: My bad, it should've been `GET` in the first place.  Fixed the path and updated the original endpoint to be `GET`.

PR #279, opened 2023-08-31
[MODDATAIMP-898] Create DI system user

ncovercash:
# [Jira MODDATAIMP-898](https://issues.folio.org/browse/MODDATAIMP-898)

## Purpose

When running jobs from the queue of split pieces, the module needs to authenticate itself.  It can not do this as the original user as this would require storing their token/credentials and runs into edge cases such as if the user is deactivated, their session ends, etc.  Therefore, a system user with all the permissions to process DI jobs is needed.

## Approach

On installation (`POST /_/tenant`), the module checks for a system user and, if it is not present, creates one.

### Testing approach
To prevent this addition from affecting all REST tests (since they all call `/_/tenant`), some special code was added to the `AbstractRestTest`, creating a special short-lived wiremock server that simulates responses for the duration of the test setup.  After the initial setup, this server is then stopped, to prevent its mocks polluting other requests.

# Test coverage: 💯 

## :sparkles: Bonus :sparkles:

- Cleaned up `/_/tenant` assertions
- Consolidated `AbstractIntegrationTest` and `AbstractRestTest` (seems `AbstractRestTest` was the same, just includes postgres/S3/etc).  I would've had to duplicate the mocks for this user's creation in both, so I decided to instead consolidate the two, particularly as only one test required the `AbstractIntegrationTest`.
- Update WireMock to v3.0.0
- Add missing mocks to `StorageCleanupServiceImplTest`
- 100% coverage on these changes
- Code smells are unrelated to these changes


ncovercash:
The remaining code smell is a TODO from another ticket, and is fixed in [MODDATAIMP-836](https://issues.folio.org/browse/MODDATAIMP-836)

okolawole-ebsco:
How will existing tenants be handled?

README.md
```diff
@@ -165,6 +166,21 @@ This covers the following environment variables:
 
 For information on what these mean and how to configure them, please see [this wiki page](https://wiki.folio.org/display/FOLIOtips/Queue+and+chunk+processing+customization).
 
+## System user
+
+To enable asynchronous job launching (as part of the file splitting process), the module creates
+a system user upon installation.  By default, the system user is named `data-import-system-user`,
```
okolawole-ebsco, line 172: The standard for other modules is to have a system user specific to the module. The name of the system user points to Data Import in general not mod-data-import. So I think we should follow the existing scheme for system users so that we know which system user belongs to which module.

src/main/java/org/folio/rest/impl/util/BufferMapper.java
```diff
@@ -3,26 +3,46 @@
...
+import lombok.experimental.UtilityClass;
 
 /**
  * Util class which maps {@link Buffer} content to instance of specified type.
  */
+@UtilityClass
 public class BufferMapper {
 
   /**
    * Returns instantiated T entity object from a buffer content.
```
okolawole-ebsco, line 15: nitpick: add "asynchronous" to mirror the sync method

src/main/java/org/folio/rest/impl/util/BufferMapper.java
```diff
@@ -3,26 +3,46 @@
...
    *
    * @param buffer     buffer
    * @param entityType type of entity which will be created
-   * @return instantiated T entity object from a buffer content
+   * @return a future of instantiated T entity object from a buffer content
    */
-  public static <T> Future<T> mapBufferContentToEntity(Buffer buffer, Class<T> entityType) {
+  public static <T> Future<T> mapBufferContentToEntity(
```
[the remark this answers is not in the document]
okolawole-ebsco: Looking again this method might not be async and just return a value wrapped with a Future. `Future.succeededFuture(mapBufferContentToEntitySync(buffer, entityType))` might suffice actually.

src/main/java/org/folio/service/auth/ApiClient.java
```diff
@@ -0,0 +1,184 @@
...
+import org.apache.http.client.utils.URIBuilder;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.message.BasicNameValuePair;
+import org.apache.http.util.EntityUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.folio.dataimport.util.OkapiConnectionParams;
+
+public abstract class ApiClient {
```
okolawole-ebsco, line 38: Did you copy this class from somewhere? or write it yourself? I ask cause i think there might be something in RMB/elsewhere that does the same thing.

src/main/java/org/folio/service/auth/SystemUserAuthService.java
```diff
@@ -0,0 +1,287 @@
...
+import org.springframework.stereotype.Service;
+
+@Service
+public class SystemUserAuthService {
+
+  private static final Logger LOGGER = LogManager.getLogger();
+
+  public static final List<String> PERMISSIONS = Collections.unmodifiableList(
+    Arrays.asList(
+      "change-manager.jobexecutions.get",
```
okolawole-ebsco, line 28: Thinking this list of permissions should be in a resource file

ncovercash:
> How will existing tenants be handled?

Not entirely sure what you're asking; I think you mean for tenants with mod-data-import already installed?  The call to `/_/tenant` occurs anytime a module is updated, so this logic will be called and the user created.  I took a lot of inspiration from mod-data-export-spring, which has a similar process.

README.md
ncovercash: What would you suggest?  Some copy the module names, while others (data export, pubsub, etc) don't:

image

src/main/java/org/folio/service/auth/ApiClient.java
ncovercash: I got the idea from @TarasSpashchenko , who pointed me to [similar code in mod-oai-pmh](https://github.com/folio-org/mod-oai-pmh/blob/master/src/main/java/org/folio/oaipmh/helpers/client/InventoryClient.java).  When we were discussing, there were two other alternatives I found:
- The builtin RMB method would be to alter the other packages (users, perms, etc) to have them generate and expose additional code/packages, such as how mod-source-record-manager does it.  This would require relatively significant breaking changes in the other modules and was deemed far too much.
- The other common way I've seen in FOLIO is using something like Feign, which is part of the Spring Way.  However, retrofitting this existing RMB/Vert.x module with all the required dependencies and infrastructure for Feign would have been a major investment and also require some higher-level changes within DI.

We decided adding basic clients like this was sufficiently straighforward without major repercussions.

src/main/java/org/folio/service/auth/SystemUserAuthService.java
ncovercash: Yeah, I'm not super happy with it there either.  I'll look at alternatives.

ncovercash: Moved into an external file

src/main/java/org/folio/service/auth/ApiClient.java
okolawole-ebsco: sounds good.

README.md
okolawole-ebsco: "mod-data-import", otherwise you can leave it as is since there is no consistency with other modules.

okolawole-ebsco:
> Not entirely sure what you're asking; I think you mean for tenants with mod-data-import already installed? The call to /_/tenant occurs anytime a module is updated, so this logic will be called and the user created. I took a lot of inspiration from mod-data-export-spring, which has a similar process.

@ncovercash  I thought `/_/tenant` is called when a module is initialized for a tenant, so i was asking about existing tenants that have already been initialized. If the approach works for tenants already initialized, then all is well.

okolawole-ebsco (approved):

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -37,12 +33,16 @@ public class ModTenantAPI extends TenantAPI {
...
+  @Autowired
+  private SystemUserAuthService systemUserAuthService;
+
+  public ModTenantAPI() {
     SpringContextUtil.autowireDependencies(this, Vertx.currentContext());
   }
 
   @Override
   Future<Integer> loadData(TenantAttributes attributes, String tenantId, Map<String, String> headers, Context context) {
+    systemUserAuthService.initializeSystemUser(headers);
```
RuslanLavrov, line 45: it seems that systemUserAuthService.initializeSystemUser(headers) should be invoked using the vertx.executeBlocking() method in order not to block event loop thread since systemUserAuthService service uses a blocking http client. 
Why it was decided to use this http client instead of http client provided by vertx?

PR #280, opened 2023-09-01
[MODDATAIMP-900] Move and improve assemble endpoint

ncovercash:
# [Jira MODDATAIMP-900](https://issues.folio.org/browse/MODDATAIMP-900)

## Purpose

Assembling a S3 file is analogous to uploading a file to mod-DI in the pre-S3 interface.  Therefore, we should do the same actions (marking the `FileDefinition` as uploaded, with a source path, etc)...

## Approach

To get the upload and file definition information, we moved the `assembleStorageFile` endpoint under `/uploadDefinitions/.../files/...`; after this, it was only a matter of calling the existing `afterFileSave` method.

## :sparkles: Bonus :sparkles:

- When checking coverage of S3 methods, I noticed the `downloadUrl` endpoint was missing tests, so I:
  - Added `404` response for when the information is not found, rather than a `500`.
  - Added full test suite with positive and negative cases
- Removed unnecessary invocation of Mockito in upload URL test suite

# Test coverage: 💯

+ some extra due to improvements to added download/upload url cases

## Notes

- ~~Commits through `72e386b` are part of #279 until that one gets merged in, since that PR contains some changes for WireMock I wanted to potentially use here.~~
- The try/catch around the main `DataImportImpl` logic is redundant as all potentially exceptional logic is handled inside futures, which have their own failure path.  Did the same update to download/upload URL endpoints, as I was looking at all S3 stuff.


okolawole-ebsco (approved):

descriptors/ModuleDescriptor-template.json
```diff
@@ -129,6 +120,20 @@
...
+          "methods": [
+            "POST"
+          ],
+          "pathPattern": "/data-import/uploadDefinitions/{uploadDefinitionId}/files/{fileId}/assembleStorageFile",
+          "permissionsRequired": [
+            "data-import.assembleStorageFile.post"
+          ],
+          "modulePermissions": [
+            "configuration.entries.collection.get",
+            "change-manager.jobexecutions.post",
```
RuslanLavrov, line 133: could you say please do we need to provide this permission (necessary for jobExecution creation) for this endpoint?

ncovercash: good catch, thanks!

ncovercash: We don't need the `post` one, good catch, thanks!

RuslanLavrov: understood, thanks! "configuration.entries.collection.get" seems is not needed as well, right?

ncovercash: @RuslanLavrov fixed!

RuslanLavrov (approved):

descriptors/ModuleDescriptor-template.json
RuslanLavrov: thanks!

PR #281, opened 2023-09-01
[MODDATAIMP-893] Cancel a running DI job

evaluk:
# [Jira MODDATAIMP-893](https://issues.folio.org/browse/MODDATAIMP-893)

## Purpose

Create an API to cancel job executions and remove queue items given a parent job execution ID

## Approach

It issues database commands to and requests the change manager client remove items that are children of a given job execution id.

#### TODOS and Open Questions

- Api returns 500 on error, which may not be desirable.  Possibly change to 4XX instead

## Learning

- Unit tests were required to directly create queue items for testing.

## Notes

- With this PR, Liquibase works in the integration test environment
- `SplitFileProcessingService` tests were removed as they were broken and severely modified in another PR, so no reason to waste time fixing them here.
  - The added utility method `verifyOkStatus` needs tests.  Also, more tests for the DAO would be good now that we can use the DB


descriptors/ModuleDescriptor-template.json
```diff
@@ -16,6 +16,15 @@
       "id": "data-import",
       "version": "3.0",
       "handlers": [
+      	{
+          "methods": [
+            "DELETE"
+          ],
+          "pathPattern": "/data-import/jobexecutions/{jobExecutionId}/cancel",
```
okolawole-ebsco, line 23: seems to me that this endpoint should be `DELETE /data-import/jobexecutions/{jobExecutionId}` or `POST /data-import/jobexecutions/{jobExecutionId}/cancel`. What do you think?

okolawole-ebsco: OR `PUT /data-import/jobexecutions/{jobExecutionId}/cancel`

ncovercash: Leaning towards the `POST /cancel`, going to confer with FE

src/main/java/org/folio/rest/impl/util/BufferMapper.java
```diff
@@ -3,20 +3,25 @@
...
   /**
    * Returns instantiated T entity object from a buffer content.
    *
    * @param buffer     buffer
    * @param entityType type of entity which will be created
-   * @return instantiated T entity object from a buffer content
+   * @return a future of instantiated T entity object from a buffer content
    */
-  public static <T> Future<T> mapBufferContentToEntity(Buffer buffer, Class<T> entityType) {
+  public static <T> Future<T> mapBufferContentToEntity(
```
okolawole-ebsco, line 21: @ncovercash You made similar changes in another PR correct?

descriptors/ModuleDescriptor-template.json
ncovercash: We're going to keep it the same for consistency with previous endpoints used in mod-srm.

okolawole-ebsco (approved):

PR #282, opened 2023-09-05
[MODDATAIMP-836] Create entry point for S3 file processing

ncovercash:
# [Jira MODDATAIMP-836](https://issues.folio.org/browse/MODDATAIMP-836)

## Purpose
Create a version of the main `/processFiles` entry point which is used when file-splitting is enabled.  This entrey point accomplishes the following:
- splitting the files
- registering the appropriate composite jobs with mod-srm
- entering each slice into the DI queue

## Approach
Adds/adapts methods in the existing `SplitFileProcessingService` to perform the desired tasks.

#### TODOS and Open Questions
- [ ] It seems like `ModTenantApiTest` is redundant, as `AbstractRestTest` already provides assertions for the `/_/tenant` endpoint.  Perhaps it could be removed?

## Bonus
- removed an unused dependency
- cleaned up some of the other DITF-added integration test class names for consistency
- improved test infrastructure and layout for other parts of `SplitFileProcessingService`
- clear queue database and S3 storage between tests

### DIFF
--- a/src/main/java/org/folio/service/file/SplitFileProcessingService.java
+++ b/src/main/java/org/folio/service/file/SplitFileProcessingService.java
@@ -24,6 +24,7 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 import lombok.With;
+import org.apache.http.HttpStatus;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.dao.DataImportQueueItemDao;
@@ -90,6 +91,7 @@ public class SplitFileProcessingService {
     this.fileProcessor = fileProcessor;
   }
 
+  /** Start a job based on information passed to the /processFiles endpoint */
   public Future<Void> startJob(
     ProcessFilesRqDto entity,
     ChangeManagerClient client,
@@ -119,7 +121,8 @@ public class SplitFileProcessingService {
         )
       )
       .andThen(v -> LOGGER.info("Job split and queued successfully!"))
-      .compose(v -> Future.succeededFuture());
+      .onFailure(err -> LOGGER.error("Unable to start job: ", err))
+      .mapEmpty();
   }
 
   /** Split file and create parent job executions for a new job */
@@ -159,7 +162,8 @@ public class SplitFileProcessingService {
           );
 
         return splitInformation;
-      });
+      })
+      .onFailure(e -> LOGGER.error("Unable to initialize parent job: ", e));
   }
 
   /** Register split file parts and fill split job execution job profile/status */
@@ -214,6 +218,7 @@ public class SplitFileProcessingService {
         }
         return result;
       })
+      .onFailure(err -> LOGGER.error("Unable to initialize children", err))
       .<Void>mapEmpty()
       .andThen(v ->
         LOGGER.info("Created child job executions for {}", splitInfo.getKey())
@@ -284,6 +289,13 @@ public class SplitFileProcessingService {
               // we don't want the queue item, just the execution, so we discard the result
               // and return the execution from the earlier step
               .map(v -> execution)
+          )
+          .onFailure(err ->
+            LOGGER.error(
+              "Unable to register split file execution for {}: ",
+              key,
+              err
+            )
           )
       );
 
@@ -486,12 +498,18 @@ public class SplitFileProcessingService {
   }
 
   protected Buffer verifyOkStatus(HttpResponse<Buffer> response) {
-    if (response.statusCode() >= 200 && response.statusCode() <= 299) {
+    if (
+      response.statusCode() >= HttpStatus.SC_OK &&
+      response.statusCode() <= HttpStatus.SC_NO_CONTENT
+    ) {
       return response.bodyAsBuffer();
     } else {
       throw LOGGER.throwing(
         new IllegalStateException(
-          "Response came back with status code " + response.statusCode()
+          "Response came back with status code " +
+          response.statusCode() +
+          " and body " +
+          response.bodyAsString()
         )
       );
     }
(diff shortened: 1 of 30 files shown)


src/main/java/org/folio/service/processing/split/FileSplitService.java
```diff
@@ -114,18 +103,29 @@ public Future<CompositeFuture> splitStream(
...
-      );
+      // this composite future resolves once all are uploaded
+      .compose(cf -> cf)
+      // now let's turn this back into a List<String>
+      .map(cf -> cf.list())
+      .map(list ->
+        list.stream().map(String.class::cast).collect(Collectors.toList())
+      )
+      // and since we're all done, we can delete the temporary folder
+      .map((List<String> innerResult) -> {
```
okolawole-ebsco, line 116: nitpick: seems like compose would be appropriate instead of map. Only because of  business logic to delete folders.

src/main/java/org/folio/service/file/SplitFileProcessingService.java
okolawole-ebsco, line 1: In general, It seems there aren't many failure conditions on the chains of futures in this file. I will be prudent to have `.onFailure` calls at different points in the chain.

src/main/java/org/folio/service/file/SplitFileProcessingService.java
```diff
@@ -227,17 +391,136 @@ public Future<Void> cancelJob(
...
+      )
+      .compose(result ->
+        fileSplitService
+          .splitFileFromS3(vertx.getOrCreateContext(), key)
+          .map(result::withSplitKeys)
+      );
+  }
+
+  protected Buffer verifyOkStatus(HttpResponse<Buffer> response) {
+    if (response.statusCode() >= 200 && response.statusCode() <= 299) {
```
okolawole-ebsco, line 489: nitpick: use the numerous enums available that represent HTTP statuses instead of ints

src/main/java/org/folio/service/file/SplitFileProcessingService.java
```diff
@@ -227,17 +391,136 @@ public Future<Void> cancelJob(
...
+      return response.bodyAsBuffer();
     } else {
-      return Future.failedFuture(
-        LOGGER.throwing(
-          new IllegalStateException(
-            "Response came back with status code " + response.statusCode()
-          )
+      throw LOGGER.throwing(
```
okolawole-ebsco, line 492: Cool! Didn't know this api exists.

src/main/java/org/folio/service/processing/split/FileSplitService.java
ncovercash: The Java File API is synchronous; compose is for futures/async logic

src/main/java/org/folio/service/file/SplitFileProcessingService.java
ncovercash: I just learned about it for this PR! It [logs at the error level](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Logger.html#throwing-T-) and returns it so you can throw and log at the same time

src/main/java/org/folio/service/processing/split/FileSplitService.java
okolawole-ebsco: ah, thought this was vertx.

okolawole-ebsco (approved):

JohnC-80 (approved):

src/main/java/org/folio/service/file/SplitFileProcessingService.java
```diff
@@ -39,16 +61,168 @@ public class SplitFileProcessingService {
...
+        uploadDefinitionService.updateBlocking(
+          entity.getUploadDefinition().getId(),
+          definition ->
+            Future.succeededFuture(
+              definition.withStatus(UploadDefinition.Status.COMPLETED)
+            ),
+          params.getTenantId()
+        )
+      )
+      .andThen(v -> LOGGER.info("Job split and queued successfully!"))
```
RuslanLavrov, line 123: would be helpful to use onSuccess() method. With andThen() method the logger will be invoked even if the current future fails and as a result, the success message will appear along with the error message in the log in case of failures, which in turn may confuse during logs investigation.

src/main/java/org/folio/service/file/SplitFileProcessingService.java
```diff
@@ -39,16 +61,168 @@ public class SplitFileProcessingService {
...
+      )
+      .map((Boolean result) -> {
+        if (Boolean.FALSE.equals(result)) {
+          throw new IllegalStateException("Could not mark job as in progress");
+        }
+        return result;
+      })
+      .onFailure(err -> LOGGER.error("Unable to initialize children", err))
+      .<Void>mapEmpty()
+      .andThen(v ->
```
RuslanLavrov, line 223: here as well

src/main/java/org/folio/service/processing/split/FileSplitService.java
```diff
@@ -114,18 +103,29 @@ public Future<CompositeFuture> splitStream(
...
+        LOGGER.info("Deleting temporary folder={}", tempDir);
+
+        try {
+          Files.delete(tempDir);
```
RuslanLavrov, line 120: it should be invoked within vertx.executeBlocking() same to minioStorageService.write() is used here in the FileSplitWriter class, or alternatively vertx fileSystem instance can be applied for non-blocking deletion

src/main/java/org/folio/service/file/SplitFileProcessingService.java
ncovercash: ah whoops, I thought `andThen` was a success-only handler. will fix in a separate PR

src/main/java/org/folio/service/processing/split/FileSplitService.java
RuslanLavrov: thanks!

PR #285, opened 2023-09-12
[MODDATAIMP-895,MODDATAIMP-928] Asynchronous background worker to launch S3-based jobs

ncovercash:
# [Jira MODDATAIMP-895](https://issues.folio.org/browse/MODDATAIMP-895)
# [Jira MODDATAIMP-928](https://issues.folio.org/browse/MODDATAIMP-928)

## Purpose
Create a background set of workers to pull from the DB and run jobs that have been entered into the queue.

## Approach
This leverages a majority of the existing process APIs by wrapping them with queue/S3 logic.  Jobs are fetched from the queue every so often (configurable by env), downloaded from S3, and passed to the pre-existing processor methods for further processing.

## Coverage: :100:


src/main/java/org/folio/service/file/S3JobRunningVerticle.java
```diff
@@ -0,0 +1,329 @@
...
+    this.pollInterval = pollInterval;
+
+    this.fileProcessor = fileProcessor;
+  }
+
+  @Override
+  public void start() {
+    LOGGER.info("Running S3JobRunningVerticle");
+
+    this.pollForJobs();
```
okolawole-ebsco, line 93: consider using vertx.setPeriodic and calling this method. Saving the future for the operation. the future can be checked to skip the poll if one "job" is currently running. This may avoid your issue is stack overflow on multiple repeats and not needing a synchronized method. Here is an example: https://github.com/folio-org/mod-source-record-storage/blob/master/mod-source-record-storage-server/src/main/java/org/folio/verticle/MarcIndexersVersionDeletionVerticle.java

src/main/java/org/folio/service/file/S3JobRunningVerticle.java
```diff
@@ -0,0 +1,329 @@
...
+          return job;
+        } catch (IOException e) {
+          throw new UncheckedIOException(e);
+        }
+      });
+  }
+
+  protected File createLocalFile(DataImportQueueItem queueItem) {
+    try {
+      File localFile = Files
```
okolawole-ebsco, line 261: Use vertx.fileSystem() to create files or run this on a worker thread to not block the event loop.

src/main/java/org/folio/service/file/S3JobRunningVerticle.java
```diff
@@ -0,0 +1,329 @@
...
+
+  /**
+   * Authenticate and get connection parameters (Okapi URL/token)
+   */
+  protected OkapiConnectionParams getConnectionParams(
+    DataImportQueueItem queueItem
+  ) {
+    OkapiConnectionParams provisionalParams = new OkapiConnectionParams(
+      Map.of(
+        "x-okapi-url",
```
okolawole-ebsco, line 288: consider using org.folio.okapi.common.XOkapiHeaders#URL and others to replace the magic strings contained in this file.

src/main/java/org/folio/service/file/S3JobRunningVerticle.java
```diff
@@ -0,0 +1,329 @@
...
+        queueItem.getTenant(),
+        "x-okapi-token",
+        token
+      ),
+      vertx
+    );
+  }
+
+  @Override
+  public void stop() {
```
okolawole-ebsco, line 315: since pollForJobs calls itself recursively, I don't think the pollForJobs process will be killed gracefully if the verticle is stopped. If so, we should kill created timers appropriately.

ncovercash: Investigated this; Vert.x cancels timers upon shutdown and, since they're invoked from a central system, there's no danger of a stack overflow

src/main/java/org/folio/service/file/S3JobRunningVerticle.java
ncovercash: Wanted to have different length delays between runs, depending on if there's work being done or not (to ensure we get work done quickly when it's there, but not hammer the DB while idle)

src/main/java/org/folio/service/file/S3JobRunningVerticle.java
ncovercash: There shouldn't be any context in which that is possible? The verticle is started as the module starts and ended only at shutdown

src/main/java/org/folio/service/file/S3JobRunningVerticle.java
ncovercash: Many changes made to this; will be ready for review later today (hopefully)

TarasSpashchenko (approved):

src/main/java/org/folio/service/file/S3JobRunningVerticle.java
```diff
@@ -0,0 +1,330 @@
...
+          queueItem.getJobExecutionId()
+        )
+      )
+      .onComplete((AsyncResult<QueueJob> v) -> {
+        queueItemDao.deleteDataImportQueueItem(queueItem.getId());
+
+        File file = localFile.get();
+        if (file != null) {
+          try {
+            FileUtils.delete(file);
```
RuslanLavrov, line 215: can vertx fileSystem instance be used for deletion here as well, just like it is used for local file creation?

ncovercash: Just added this, thanks!

RuslanLavrov (approved):

PR #289, opened 2023-09-14
[MODDATAIMP-914] Prefix S3 files with data-import/

ncovercash:
# [Jira MODDATAIMP-914](https://issues.folio.org/browse/MODDATAIMP-914)

## Purpose

- Prefix S3 file operations with `data-import/` to improve usability within multi-module buckets

## Approach

- Added checks for this prefix in S3 calls to prevent potential abuse

## Bonus
- Cleaned up MinIO code to remove redundant future returns (can return `executeBlocking` directly as its `Future` form)
  - I also cleaned up the `completeMultipartFileUpload` method to fail on error, rather than succeed with `false`.
  - A majority of the changed lines are due to this as well as cleanup in the associated tests

## Test coverage :100:

### DIFF
--- a/src/main/java/org/folio/service/s3storage/MinioStorageServiceImpl.java
+++ b/src/main/java/org/folio/service/s3storage/MinioStorageServiceImpl.java
@@ -1,6 +1,5 @@
 package org.folio.service.s3storage;
 
-import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.Vertx;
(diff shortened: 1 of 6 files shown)


RuslanLavrov (approved):

Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/service/s3storage/MinioStorageServiceImpl.java
```diff
@@ -91,160 +80,115 @@ public Future<FileUploadInfo> getFileUploadPartUrl(
...
-        }
+    return vertx.executeBlocking((Promise<InputStream> blockingFuture) -> {
+      try {
+        verifyKey(key);
+
+        LOGGER.info("Created input stream to read remote file for key {}", key);
+        InputStream inStream = client.read(key);
+        blockingFuture.complete(inStream);
+      } catch (S3ClientException e) {
+        LOGGER.error("Could not read from S3:", e);
```
VRohach, line 119: Is there an opportunity to log something else? I mean filename or some other identifier?

ncovercash: It should contain detailed information in the exception; additionally, it logs right before `Created input stream to read remote file for key {}`, which would contain the identifier.

PR #291, opened 2023-09-18
[MODDATAIMP-921] Handle non-MARC files properly

ncovercash:
# [Jira MODDATAIMP-921](https://issues.folio.org/browse/MODDATAIMP-921)

## Purpose
Currently, all files were fed through the file-splitting process.  However, this system only supports plain binary MARC files (.mrc); as such, MARC JSON, MARC XML, and EDIFACT records were being misprocessed.

Futhermore, the logic used to calculate the number of records in the file was also assuming its type as binary MARC (.mrc).  This changes this logic to use already implemented logic from the file processing components.

## Approach
This adds utility methods and checks to ensure only binary MARC files (.mrc) are split; all other file types are fed through an alternative workflow.  Additionally, the previously-used MARC record counting was adapted to work with any supported type.

## Coverage: :100:

### DIFF
--- a/src/main/java/org/folio/service/file/SplitFileProcessingService.java
+++ b/src/main/java/org/folio/service/file/SplitFileProcessingService.java
@@ -136,7 +136,7 @@ public class SplitFileProcessingService {
         .getFileDefinitions()
         .stream()
         .map(FileDefinition::getSourcePath)
-        .map(this::splitFile)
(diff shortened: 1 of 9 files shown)


ncovercash:
Not sure where Sonar's comment is:

TarasSpashchenko (approved):

KaterynaSenchenko (approved):

PR #292, opened 2023-09-19
MODCON-92. Populate system users with type='system' in different modules

SerhiiNosko:
## Purpose
https://issues.folio.org/browse/MODCON-92

## Approach
User type is required now in ECS mode, so each system user should have type='system' populated


ncovercash (approved):
Thanks!

Dmitriy-Butramyou (approved):

okolawole-ebsco (approved):

PR #306, opened 2023-10-13
release 3.0.0

KaterynaSenchenko:


NEWS.md
```diff
@@ -1,4 +1,4 @@
-## 2023-10-04 v2.8.0
```
ncovercash, line 1: Can we keep both in here, as one is for orchid and the other poppy? I don't want future readers to be unable to find v2.8 changes

ncovercash (changes requested):

NEWS.md
KaterynaSenchenko: Thanks, I added a section for 2.7.2 (Orchid), but for Poppy we are bumping the version to 3.0.0

PR #308, opened 2023-10-23
[MODDATAIMP-947] Fix improper system user account management

ncovercash:
# [Jira MODDATAIMP-947](https://issues.folio.org/browse/MODDATAIMP-947)

## Purpose
With Poppy, certain changes were made to user logins/token management.  We updated the code here to handle this, however, missed a couple spots relating to password setting on tenant install.

## Approach
Ensure proper handling of login results when installing on tenant

### DIFF
--- a/src/main/java/org/folio/rest/impl/ModTenantAPI.java
+++ b/src/main/java/org/folio/rest/impl/ModTenantAPI.java
@@ -46,10 +46,14 @@ public class ModTenantAPI extends TenantAPI {
 
   @Override
   Future<Integer> loadData(TenantAttributes attributes, String tenantId, Map<String, String> headers, Context context) {
-    if (fileSplittingEnabled) {
-      systemUserAuthService.initializeSystemUser(headers);
-    }
-    return super.loadData(attributes, tenantId, headers, context)
+    return Future.succeededFuture()
+      .compose(v -> {
+        if (fileSplittingEnabled) {
+          return systemUserAuthService.initializeSystemUser(headers);
+        }
+        return Future.succeededFuture();
+      })
+      .compose(v -> super.loadData(attributes, tenantId, headers, context))
       .compose(num -> {
         initStorageCleanupService(headers, context);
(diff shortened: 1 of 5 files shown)


KaterynaSenchenko (approved):

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -46,10 +46,14 @@ public ModTenantAPI() {
...
   @Override
   Future<Integer> loadData(TenantAttributes attributes, String tenantId, Map<String, String> headers, Context context) {
-    if (fileSplittingEnabled) {
-      systemUserAuthService.initializeSystemUser(headers);
-    }
-    return super.loadData(attributes, tenantId, headers, context)
+    return Future.succeededFuture()
+      .compose(v -> {
+        if (fileSplittingEnabled) {
+          return systemUserAuthService.initializeSystemUser(headers);
```
okolawole-ebsco, line 52: I think the system user should be created everytime. If a tenant is installed with file splitting disabled, it cannot have DI working when file splitting is enabled.

RomanChernetskyi (approved):

src/main/java/org/folio/rest/impl/ModTenantAPI.java
ncovercash: Our goal was to not have file splitting be a breaking change/only opt-in; IMO having to set its password environment variable is a breaking change.

I'll add a note to the release notes though; in generally, I'd advise calling install whenever changing configuration like this.  The system user's password env variable would have the same issue, where an install is required each time

ncovercash: Also, I don't want to spook admins with a new user appearing out of the blue

okolawole-ebsco: Sounds good!

PR #309, opened 2023-10-23
[MODDATAIMP-942] [Poppy] Add missing permissions

ncovercash:
# [Jira MODDATAIMP-942](https://issues.folio.org/browse/MODDATAIMP-942)

Poppy version of #310

## Purpose
A couple straggling permissions needed to be added for full DI functionality:

- Needed for EDITFACT imports with acquisition units (see FAT-1470)
  - invoices.acquisitions-units-assignments.assign
  - invoices.acquisitions-units-assignments.manage

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -255,7 +255,10 @@
             "acquisitions-units-storage.memberships.collection.get",
             "isbn-utils.convert-to-13.get",
             "instance-authority-links.instances.collection.get",
-            "instance-authority-links.instances.collection.put"
+            "instance-authority-links.instances.collection.put",
+            "instance-authority.linking-rules.collection.get",
+            "user-tenants.collection.get",
+            "organizations.organizations.collection.get"
           ],
           "permissionsDesired": [
             "invoices.acquisitions-units-assignments.assign",
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -258,7 +258,9 @@
             "instance-authority-links.instances.collection.put",
             "instance-authority.linking-rules.collection.get",
             "user-tenants.collection.get",
-            "organizations.organizations.collection.get"
+            "organizations.organizations.collection.get",
+            "invoices.acquisitions-units-assignments.assign",
+            "invoices.acquisitions-units-assignments.manage"
           ],
           "permissionsDesired": [
             "invoices.acquisitions-units-assignments.assign",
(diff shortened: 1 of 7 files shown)


src/main/resources/permissions.txt
```diff
@@ -68,8 +68,6 @@ invoice-storage.invoices.item.put
 invoice-storage.invoice-lines.item.post
 invoice-storage.invoice-lines.item.put
 invoice-storage.invoice-lines.collection.get
-acquisitions-units.units.collection.get
-acquisitions-units.memberships.collection.get
```
ncovercash, line 72: These were duplicated elsewhere in the file

KaterynaSenchenko (approved):

RomanChernetskyi (approved):

descriptors/ModuleDescriptor-template.json
```diff
@@ -258,7 +258,9 @@
             "instance-authority-links.instances.collection.put",
             "instance-authority.linking-rules.collection.get",
             "user-tenants.collection.get",
-            "organizations.organizations.collection.get"
+            "organizations.organizations.collection.get",
+            "invoices.acquisitions-units-assignments.assign",
```
RuslanLavrov, line 262: we should not add this permission here, since this permission is optional and a user assigned to a particular acquisition unit should be directly provided with this permission. Otherwise, a user who is not supposed to be provided with this permission will receive it.

KaterynaSenchenko: Can it be added to permissionsDesired? we have it in mod-srm in that section? Or is it part of the permission set that needs to be assigned to the user?

ncovercash: fixed

src/main/resources/permissions.txt
```diff
@@ -101,3 +99,6 @@ instance-authority-links.instances.collection.get
 instance-authority-links.instances.collection.put
 instance-authority.linking-rules.collection.get
 user-tenants.collection.get
+organizations.organizations.collection.get
+invoices.acquisitions-units-assignments.assign
```
RuslanLavrov, line 103: it seems that for the data-import system user we should not add this permission as well because data import processes initiated by all other users will be provided with this permission, regardless of whether a user starting the import process has this permission or not.

ncovercash: I'm not sure what to do, then, since the data import system user operates on the user's behalf to process jobs. Anything the user can do the system user must be able to do

RuslanLavrov: I haven't come to some solution, but might be we can consider approach of saving "X-Okapi-Permissions" header of the initial user along with the DataImportQueueItem (or smth like that) in the DB. This way, the header could be passed further through Kafka headers and be used by the logic that determines whether the user has this specific permission during invoice creation.

ncovercash:
See #311

PR #311, opened 2023-10-27
[MODDATAIMP-953] [Poppy] Pass user information directly into Kafka

ncovercash:
Poppy version of #312

## Purpose
Operations involving acquisition units and invoices require information about the original user to determine if they are permitted to perform actions and are members of acquisition units.

## Approach
Instead of using the system user, we store and pass the originating user's tokens, permissions, and user ID to Kafka/SRM when processing jobs.

Since this was the only time we needed to use the system user, the system user is also removed.


src/main/java/org/folio/service/file/SplitFileProcessingService.java
```diff
@@ -287,6 +287,21 @@ protected Future<Void> initializeChildren(
...
                         entity.getJobProfileInfo().getDataType().toString()
                       )
+                      .withOkapiToken(params.getToken())
+                      .withOkapiPermissions(
+                        params
+                          .getHeaders()
+                          .entries()
+                          .stream()
+                          .filter(header ->
+                            "x-okapi-permissions".equalsIgnoreCase(
```
okolawole-ebsco, line 297: nit: could use XOkapiHeaders.PERMISSIONS here

okolawole-ebsco (approved):

RuslanLavrov (approved):

KaterynaSenchenko (approved):

PR #310, opened 2023-10-23
[MODDATAIMP-942] [Orchid] Add missing permissions

ncovercash:
# [Jira MODDATAIMP-942](https://issues.folio.org/browse/MODDATAIMP-942)

Orchid version of #309 

## Purpose
A couple straggling permissions needed to be added for full DI functionality:

- Needed for EDITFACT imports with acquisition units (see FAT-1470)
  - invoices.acquisitions-units-assignments.assign
  - invoices.acquisitions-units-assignments.manage
- Part of mod-entities-links and potentially user creation/update
  - user-tenants.collection.get

PR #312, opened 2023-10-27
[MODDATAIMP-953] [Orchid] Pass user information directly into Kafka

ncovercash:
Orchid version of #311

## Purpose
Operations involving acquisition units and invoices require information about the original user to determine if they are permitted to perform actions and are members of acquisition units.

## Approach
Instead of using the system user, we store and pass the originating user's tokens, permissions, and user ID to Kafka/SRM when processing jobs.

Since this was the only time we needed to use the system user, the system user is also removed.


descriptors/ModuleDescriptor-template.json
```diff
@@ -259,6 +247,7 @@
           ],
           "permissionsDesired": [
             "invoices.acquisitions-units-assignments.assign",
+            "invoices.acquisitions-units-assignments.manage",
```
RuslanLavrov, line 250: seems there is no need to add "invoices.acquisitions-units-assignments.manage" permission here for Orchid and for Poppy branches, as far as I can see it is not used during invoice creation.
Please, add "organizations.organizations.collection.get" permission as modulePermission above (as it was added in the PR https://github.com/folio-org/mod-data-import/pull/310). Looks like it makes sense to have it in the Orchid branch as well.

ncovercash: fixed!

KaterynaSenchenko (approved):

PR #316, opened 2024-01-24
MODDATAIMP-886 Create Kafka topics instead of relying on auto create

JavokhirAbdullayev:
## Purpose
Create Kafka topics instead of relying on auto create [MODDATAIMP-886](https://issues.folio.org/browse/MODDATAIMP-886)

## Approach
* call createKafkaTopics method of KafkaAdminClientService in ModTenantApi class
* add tests


src/main/java/org/folio/rest/impl/util/DataImportKafkaTopic.java
```diff
@@ -0,0 +1,38 @@
...
+  DI_MARC_BIB_FOR_ORDER_CREATED("di-marc-bib-for-order-created"),
+  DI_MARC_FOR_UPDATE_RECEIVED("di-marc-for-update-received"),
+  DI_PARSED_RECORDS_CHUNK_SAVED("di-parsed-records-chunk-saved"),
+  DI_RAW_RECORDS_CHUNK_PARSED("di-raw-records-chunk-parsed"),
+  DI_RAW_RECORDS_CHUNK_READ("di-raw-records-chunk-read"),
+  DI_SRS_MARC_AUTHORITY_RECORD_CREATED("di-srs-marc-authority-record-created"),
+  DI_SRS_MARC_AUTHORITY_RECORD_NOT_MATCHED("di-srs-marc-authority-record-not-matched"),
+  DI_SRS_MARC_BIB_INSTANCE_HRID_SET("di-srs-marc-bib-instance-hrid-set"),
+  DI_SRS_MARC_BIB_RECORD_CREATED("di-srs-marc-bib-record-created"),
+  DI_SRS_MARC_BIB_RECORD_MODIFIED_READY_FOR_POST_PROCESSING("di-srs-marc-bib-record-modified-ready-for-post-processing");
```
KaterynaSenchenko, line 23: I'm not sure mod-data-import module should be responsible for all DI topics, I'd leave here only the topic that this particular module publishes to

JavokhirAbdullayev: removed redundant topics, only mod-di related topics left

src/main/java/org/folio/rest/impl/util/DataImportKafkaTopic.java
```diff
@@ -0,0 +1,27 @@
+package org.folio.rest.impl.util;
+
+import org.folio.kafka.services.KafkaTopic;
+
+public enum DataImportKafkaTopic implements KafkaTopic {
+  DI_COMPLETED("DI_COMPLETED"),
```
KaterynaSenchenko, line 6: mod-data-import does not publish in this topics neither. Please review the publisher, let's leave only those topics that mod-data-import needs to publish to

README.md
```diff
@@ -205,6 +205,11 @@ All modules involved in data import (mod-data-import, mod-source-record-manager,
 * `OKAPI_URL`
 * `ENV` (unique env ID)
 
+There are another important properties - `number of partitions` for topics `DI_INITIALIZATION_STARTED` and `DI_RAW_RECORDS_CHUNK_READ`
+which are created during tenant initialization, the values of which can be customized with
+ `DI_INITIALIZATION_STARTED_PARTITIONS` and `DI_RAW_RECORDS_CHUNK_READ_PARTITIONS` env variables respectively. 
+Default value - `1`.
```
mtraneis, line 211: Is there an option in the Kafka wrapper to specify the Replication Factor (number of brokers to which the data is replicated)?

JavokhirAbdullayev: hi, yes can be set via env var REPLICATION_FACTOR

mtraneis: @okolawole-ebsco and @KaterynaSenchenko , should the Replication Factor be set during topic creation as well?

okolawole-ebsco: according to what i have read so far, replication factor will be set to 1. I think that is the correct value for a default for Kafka.

mtraneis: What I meant to ask is: should the RF be configurable (if it's not configurable already) so that hosting providers can set it despite its default value can be 1?

JavokhirAbdullayev: yes, as mentioned earlier we have this env var REPLICATION_FACTOR, to be set value for topics created by this module, by default the value of RF is 1

KaterynaSenchenko (approved):

RuslanLavrov (approved):

PR #354, opened 2025-02-06
MODDATAIMP-1125 Remove dependency on mod-configuration

JavokhirAbdullayev:
## Purpose
[MODDATAIMP-1125](https://folio-org.atlassian.net/browse/MODDATAIMP-1125) Remove dependency on mod-configuration

## Approach
* remove mod-configuration's dependency

### DIFF
--- a/src/test/java/org/folio/rest/AbstractRestTest.java
+++ b/src/test/java/org/folio/rest/AbstractRestTest.java
@@ -131,7 +131,7 @@ public abstract class AbstractRestTest {
 
   @Container
   private static final LocalStackContainer localStackContainer = new LocalStackContainer(
-    DockerImageName.parse("localstack/localstack:0.11.3")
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -27,7 +26,6 @@
 
 public class ModTenantAPI extends TenantAPI {
 
-  private static final String DELAY_TIME_BETWEEN_CLEANUP_CODE = "data.import.cleanup.delay.time";
   private static final long DELAY_TIME_BETWEEN_CLEANUP_VALUE_MILLIS = 3600_000;
```
JavokhirAbdullayev, line 29: data.import.cleanup.delay.time 
data.import.cleanup.time 
pass both as env vars

KaterynaSenchenko (approved):

PR #366, opened 2025-06-13
MODDATAIMP-1214 - Add permission to permit update of shared instance and MARC

RuslanLavrov:
## Purpose
to restrict shared instance/record update through data import

## Approach
* add permission that will allow update of shared instance/record through data import

## Learning
[MODDATAIMP-1214](https://issues.folio.org/browse/MODDATAIMP-1214)


KaterynaSenchenko:
After discussion, let's put this permission declaration in the mod-consortia module. Thank you!

Aliaksandr-Fedasiuk (approved):

PR #368, opened 2025-07-31
MODDATAIMP-1223 - File splitting (DI) creating duplicate records

RuslanLavrov:
## Purpose
to fix "file contains duplicates" error that occurs when importing large files

## Approach
* fix lock acquisition and transactional operations to ensure that imported file is processed by one worker
* update tests

## Learning
[MODDATAIMP-1223](https://issues.folio.org/browse/MODDATAIMP-1223)

### DIFF
--- a/src/main/java/org/folio/dao/DataImportQueueItemDao.java
+++ b/src/main/java/org/folio/dao/DataImportQueueItemDao.java
@@ -17,15 +17,17 @@ public interface DataImportQueueItemDao {
   Future<DataImportQueueItemCollection> getAllQueueItems();
 
   /**
-   * Get all {@link DataImportQueueItem} in database where processing=false
+   * Get all {@link DataImportQueueItem} in database where processing=false using the specified {@code connection}
    *
+   * @param connection database connection
    * @return future with {@link DataImportQueueItemCollection}
    */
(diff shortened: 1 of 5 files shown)


Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/dao/DataImportQueueItemDao.java
```diff
@@ -15,24 +17,26 @@ public interface DataImportQueueItemDao {
   Future<DataImportQueueItemCollection> getAllQueueItems();
 
   /**
-   * Get all {@link DataImportQueueItem} in database where processing=false
+   * Get all {@link DataImportQueueItem} in database where processing=false using the specified {@code connection}
```
okolawole-ebsco, line 20: why is a connection needed?

RuslanLavrov: connection is needed to execute the method/query in transaction initiated within specified connection. getAllQueueItemsAndProcessAtomic method is intended to ensure that only one worker can process certain queue item atomically, and this is done by obtaining exclusive lock on queue items table. That's why the search in the queue items table should be executed within transaction that acquired the lock

PR #380, opened 2026-03-14
MODDATAIMP-1234 - Upgrade module to Vert.x 5.0

RuslanLavrov:
## Purpose
upgrade the module to Vertx 5

## Approach
* MODULE_SPECIFIC_ARGS no longer exists so "file.processing.buffer.record.charset" has been exposed as module property
* FileSystem.deleteRecursive(String path, boolean recursive) method has been deprecated and replaced with FileSystem.deleteRecursive(String path) in Vertx 5
* FileSplitWriter was refactored since WriteStream.write(Buffer data, Handler<AsyncResult<Void>> handler) method has been deprecated and replaced with WriteStream.write(Buffer data) in Vertx 5
* testDownloadPresignedKeyNotPresentWithSimilarFiles() and testDownloadPresignedKeyNotPresentWithNoMatch() tests have been updated because now after refactoring MinioStorageServiceImpl::getFileDownloadUrl method the client.getPresignedUrl() is no longer called if the key/file is not found in the S3 bucket (as there is no need to call client.getPresignedUrl() in this case)
* vertxContext.executeBlocking() method has been deleted from the FileSplitWriter::uploadChunkAsync() method, because
the minioStorageService.write method already executes logic in vertxContext.executeBlocking()
* fixed FileSplitWriterExceptionalTest::testInvalidDirectory() test - the FileSplitWriter::end() method has been updated with the check before failing `chunkUploadingCompositeFuturePromise` to prevent failing it twice, if specified `chunkFolder` does not exist
* the ensureProcessFilesRqDtoWithFileDefinitionSourcePath() method has been added to the `POST /data-import/uploadDefinitions/{uploadDefinitionId}/processFiles` endpoint logic to ensure that the `ProcessFilesRqDto` request contains the FileDefinition.sourcePath field value. Because, since RMB v36.0.0, the [FileDefinition.sourcePath field](https://github.com/folio-org/data-import-raml-storage/blob/811c0170b212d29cbd40e5eb7d6a45949e636208/schemas/mod-data-import/fileDefinition.json#L11), which is defined as a read-only field, comes as null to the endpoint logic

## Learning
[MODDATAIMP-1234](https://issues.folio.org/browse/MODDATAIMP-1234)

### DIFF
--- a/src/main/java/org/folio/rest/impl/DataImportImpl.java
+++ b/src/main/java/org/folio/rest/impl/DataImportImpl.java
@@ -281,24 +281,26 @@ public class DataImportImpl implements DataImport {
 
         ensureProcessFilesRqDtoWithFileDefinitionSourcePath(entity)
           .compose(v -> {
+            asyncResultHandler.handle(Future.succeededFuture(
+              PostDataImportUploadDefinitionsProcessFilesByUploadDefinitionIdResponse.respond204()));
+
             if (this.fileSplittingEnabled) {
               OkapiConnectionParams params = new OkapiConnectionParams(okapiHeaders, vertxContext.owner());
+              var httpClient = vertxContext.owner().createHttpClient();
               ChangeManagerClient changeManagerClient = new ChangeManagerClient(params.getOkapiUrl(),
-                params.getTenantId(), params.getToken(), vertxContext.owner().createHttpClient());
-              return splitFileProcessingService.startJob(entity, changeManagerClient, params)
-                .mapEmpty();
+                params.getTenantId(), params.getToken(), httpClient);
+              splitFileProcessingService.startJob(entity, changeManagerClient, params);
+            } else {
+              fileProcessor.process(JsonObject.mapFrom(entity), JsonObject.mapFrom(okapiHeaders));
             }
-            fileProcessor.process(JsonObject.mapFrom(entity), JsonObject.mapFrom(okapiHeaders));
             return Future.succeededFuture();
           })
-          .onFailure(e -> LOGGER.warn("postDataImportUploadDefinitionsProcessFilesByUploadDefinitionId:: Error during file processing for uploadDefinitionId: {}", uploadDefinitionId, e));
-
-        Future.succeededFuture()
-          .map(PostDataImportUploadDefinitionsProcessFilesByUploadDefinitionIdResponse.respond204())
-          .map(Response.class::cast)
-          .onComplete(asyncResultHandler);
+          .onFailure(e -> {
+            LOGGER.warn("postDataImportUploadDefinitionsProcessFilesByUploadDefinitionId:: Error during file processing for uploadDefinitionId: {}", uploadDefinitionId, e);
+            asyncResultHandler.handle(Future.succeededFuture(ExceptionHelper.mapExceptionToResponse(e)));
+          });
       } catch (Exception e) {
-        LOGGER.warn("postDataImportUploadDefinitionsProcessFilesByUploadDefinitionId:: Cannot upload definitions process files by uploadDefinitionId {}", uploadDefinitionId);
+        LOGGER.warn("postDataImportUploadDefinitionsProcessFilesByUploadDefinitionId:: Cannot upload definitions process files by uploadDefinitionId {}", uploadDefinitionId, e);
         asyncResultHandler.handle(Future.succeededFuture(ExceptionHelper.mapExceptionToResponse(e)));
       }
     });
--- a/src/main/java/org/folio/rest/impl/DataImportImpl.java
+++ b/src/main/java/org/folio/rest/impl/DataImportImpl.java
@@ -279,31 +279,24 @@ public class DataImportImpl implements DataImport {
       try {
         LOGGER.info("postDataImportUploadDefinitionsProcessFilesByUploadDefinitionId:: Starting file processing for upload definition {}", uploadDefinitionId);
 
-        if (this.fileSplittingEnabled) {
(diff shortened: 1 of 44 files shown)


KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/rest/impl/DataImportImpl.java
```diff
@@ -279,31 +279,24 @@ public void postDataImportUploadDefinitionsProcessFilesByUploadDefinitionId(Stri
...
+                params.getTenantId(), params.getToken(), vertxContext.owner().createHttpClient());
+              return splitFileProcessingService.startJob(entity, changeManagerClient, params)
+                .mapEmpty();
+            }
+            fileProcessor.process(JsonObject.mapFrom(entity), JsonObject.mapFrom(okapiHeaders));
+            return Future.succeededFuture();
+          })
+          .onFailure(e -> LOGGER.warn("postDataImportUploadDefinitionsProcessFilesByUploadDefinitionId:: Error during file processing for uploadDefinitionId: {}", uploadDefinitionId, e));
+
+        Future.succeededFuture()
```
okolawole-ebsco: Does this execute before `fileProcessor.process` or even `ensureProcessFilesRqDtoWithFileDefinitionSourcePath` completes?

If so it means we are sending HTTP 204(success) before db access and other processes complete. To clarify, we should still maintain fire and forget but that should be right before we begin processing the file.

src/main/java/org/folio/rest/impl/ModTenantAPI.java
```diff
@@ -96,14 +96,10 @@ private void initStorageCleanupService(Map<String, String> headers, Context cont
...
-        e -> vertx.executeBlocking(b -> storageCleanupService.cleanStorage(params),
-          cleanupAr -> {
-            if (cleanupAr.failed()) {
-              LOGGER.error("Error during cleaning file storage.", cleanupAr.cause());
-            } else {
-              LOGGER.info("File storage was successfully cleaned of unused files");
-            }
-      }));
+    vertx.setPeriodic(delayTimeBetweenCleanupValueMillis,
+      id -> vertx.executeBlocking(() -> storageCleanupService.cleanStorage(params))
```
okolawole-ebsco: `storageCleanupService.cleanStorage` returns a `Future<Boolean>`. vertx.executeBlocking wraps the output with a Future so the final output becomes `Future<Future<Boolean>>`. This means that the `.onSuccess` and 'onFailure` below are hooked on the outer Future of `Future<Future<Boolean>>` and will resolve immediately. The inner Future will resolve later but there is no one to listen.

This bug existed in the old code as well.

src/main/java/org/folio/service/file/SplitFileProcessingService.java
```diff
@@ -111,33 +111,22 @@ public Future<Void> startJob(
     ChangeManagerClient client,
     OkapiConnectionParams params
   ) {
-    return executor.executeBlocking(
-      promise ->
+    Promise<Void> promise = Promise.promise();
+    executor.executeBlocking(
+      () -> {
         initializeJob(entity, client)
```
okolawole-ebsco: Since this is async, running it in `executeBlocking` weird looking. We could try rewriting this without `executeBlocking`

src/main/java/org/folio/service/processing/split/FileSplitWriter.java
```diff
@@ -92,20 +91,17 @@ public WriteStream<Buffer> exceptionHandler(
   @Override
   public Future<Void> write(Buffer data) {
```
okolawole-ebsco, line 88: To stay consistent with the rest of the PR, we should eliminate the promise pattern and just return a future from `write(data, handler)`.

src/main/java/org/folio/rest/impl/DataImportImpl.java
RuslanLavrov: yes, it is called before `ensureProcessFilesRqDtoWithFileDefinitionSourcePath()`, and `fileProcessor.process(`) or `splitFileProcessingService.startJob()` (both of which perform queries to the database under the hood). The changes preserve the asynchronous behaviour of the endpoint as it was before.
Do you mean to send a response after ensureProcessFilesRqDtoWithFileDefinitionSourcePath() completion?

src/main/java/org/folio/rest/impl/ModTenantAPI.java
RuslanLavrov: updated thanks

src/main/java/org/folio/service/file/SplitFileProcessingService.java
RuslanLavrov: updated and went through subsequent methods that are called by startJob() method to apply executeBlocking more granularly (only for blocking operations) and to delete it where it's redundant

src/main/java/org/folio/service/processing/split/FileSplitWriter.java
RuslanLavrov: updated

src/main/java/org/folio/rest/impl/DataImportImpl.java
okolawole-ebsco: Yes, I think we should send a response after ensureProcessFilesRqDtoWithFileDefinitionSourcePath() completion.

RuslanLavrov: updated

